PR #275, opened 2014-05-26
advance network performance,using customize threadpool replacing the def...

absurdliyang:
...ault threadpool which alive num is 1 in AsyncTask to fix the network block when many asyctask is executing.


paulbatum:
Closing due to inactivity.

PR #398, opened 2014-09-18
Xamarin.iOS Unified API Support

Redth:
This is at least a start to adding Xamarin.iOS Unified API support to the Xamarin Component.

- XamarinVS 3.6.x is required (in the Beta channel as of writing)
- NuGet 2.8.3 Alpha is required to be installed (just released on nuget.codeplex.com)

Basically, the `iOS.csproj` is now the Unified API (this should be the default going forward).  There is a copy named `iOS-Classic.csproj` which is for the old, unchanged classic API.  `component.yaml` has been updated to reflect the new `ios-unified` target.  The ***Xamarin.Auth*** submodule has also been updated to include its update for Unified API support.

I may be missing some things, I think there may be tooling to build this that is not in the repo.  Also I did not port the Sample app to Unified (we can leave it as a Classic API sample for now, and eventually we should just convert it and drop the classic sample going forward).  I didn't port the Tests either, you may want to do that.

The important thing for now is to get the component updated with Unified support so users of these new project types are not left in the dark!

Thanks!


paulbatum:
Latest version of this work is tracked under pull request https://github.com/Azure/azure-mobile-services/pull/545

paulbatum:
Version 1.0.1 of the WindowsAzure.MobileServices.SQLiteStore package and version 1.3.1 of the WindowsAzure.MobileServices package both support the unified API.

PR #550, opened 2015-01-09
adding a force-purge option for deleting offline records, operations, an...

brettsam:
...d metadata


sdk/iOS/src/MSQueuePurgeOperation.m
```diff
@@ -0,0 +1,194 @@
...
+    [self didChangeValueForKey:@"isFinished"];
+}
+
+// Check if the operation was cancelled and if so, begin winding down
+-(BOOL) checkIsCanceled
+{
+    if (self.isCancelled) {
+        NSError *error = [self errorWithDescription:@"Purge cancelled" code:MSPullAbortedUnknown];
+        [self callCompletionIfError:error];
+        [self completeOperation];
```
phvannor: Remove, callCompletion would already have called completeOperation

phvannor:
:+1:

Issue #491, opened 2014-11-14
InsertAsync issues on MobileServiceSyncTable<T>


jcmm33:
It would appear that performing an Insert with a record that currently exists doesn't result in an exception, but instead seems to perform an Upsert the first time it is tried.

Then trying the same insert again results in an exception 
System.InvalidOperationException: An insert operation on the item is already in the queue.

I would expect the first insert to fail.

So, in this sample below where there is already a db record with Id="2" we get the first two Debug lines output with values of 1 for the count, but the second insertasync then causes the exception to be raised.

            var record = new Simple() {Id = "2", Name = System.DateTime.Now.ToLongTimeString()};

            var existing = (await this.simpleTable.Where(u => u.Id == record.Id).ToListAsync());

            Debug.WriteLine("No Records #Pass 1 {0}:",existing.Count);

            await this.simpleTable.InsertAsync(record);

            existing = (await this.simpleTable.Where(u => u.Id == record.Id).ToListAsync());

            Debug.WriteLine("No Records #Pass 2 {0}:", existing.Count);

            await this.simpleTable.InsertAsync(record);

            existing = (await this.simpleTable.Where(u => u.Id == record.Id).ToListAsync());

            Debug.WriteLine("No Records #Pass 3 {0}:", existing.Count);

phvannor:
Fix complete for iOS, Android needs investigated/fixed if applicable.

soninaren:
Investigated this with android. It needs to be fixed. 
Steps:
1) Insert a record into the mobile service database
2) Pull the record to the local store
3) Insert a duplicate record - no exception received 
4) Insert the same record again - Exception is thrown

marianosz:
@ysxu Mimi, this bug is already fixed on pull request #560

PR #552, opened 2015-01-12
insert returns error if item already exists

brettsam:


phvannor:
This address #491 for iOS

PR #556, opened 2015-01-15
[iOS] Helper function for NSManagedObject

phvannor:
Expose a helper function on MSCoreDataStore that makes it simple to go
from a NSManagedObject back to the expected dictionary format.


brettsam:
Looks good.

Issue #534, opened 2014-12-18
SQLite error in table __errors in Android 2.0.1-beta 


lordjesus:
I am receiving an error when pushing data from Android to the cloud. I suspect this has something to do with the LocalStore implementation. When I look into the db file on the Android device, there is no column named operationkind in the __errors table. The weird thing is that I have two devices both running Android 5, running the exact same build of the app, and only one of the devices show this error, and thus, refuses to sync data.

12-18 12:47:39.740    6336-6963/com.dkh.app.azureinspection E/SQLiteLog﹕ (1) table __errors has no column named operationkind
12-18 12:47:39.745    6336-6363/com.dkh.app.azureinspection W/System.err﹕ java.util.concurrent.ExecutionException: com.microsoft.windowsazure.mobileservices.table.sync.push.MobileServicePushFailedException
12-18 12:47:39.745    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at com.google.common.util.concurrent.AbstractFuture$Sync.getValue(AbstractFuture.java:299)
12-18 12:47:39.745    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:286)
12-18 12:47:39.745    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:116)
12-18 12:47:39.746    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at com.dkh.app.azureinspection.INSTAActivity$1.doInBackground(INSTAActivity.java:137)
12-18 12:47:39.746    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at com.dkh.app.azureinspection.INSTAActivity$1.doInBackground(INSTAActivity.java:130)
12-18 12:47:39.746    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at android.os.AsyncTask$2.call(AsyncTask.java:288)
12-18 12:47:39.746    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at java.util.concurrent.FutureTask.run(FutureTask.java:237)
12-18 12:47:39.746    6336-6363/com.dkh.app.azureinspection W/System.err﹕ at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
[...]

Issue #536, opened 2014-12-23
Android SDK thinks virtual fields from backend are missing columns


lordjesus:
This issue might be related to issue #534 and #505.

I have a table on my .NET backend server

    public class INSTA : EntityData
    {
        public string InspectionId { get; set; }
        public virtual Inspection Inspection { get; set; }

        public string LevelsInInspectionId { get; set; }
        public virtual LevelsInInspection LevelsInInspection { get; set; }

        public DateTime? InspectedDateTime { get; set; } // Nullable datetime
        public string Inventory { get; set; }
        public string Wall { get; set; }
        public string Floor { get; set; }
        public string Ceiling { get; set; }
        public bool Inspected { get; set; }
        public bool Rejected { get; set; }
    }

which corresponds to an SQL table 
![image](https://cloud.githubusercontent.com/assets/3789478/5537213/fc17bff0-8a9d-11e4-9a5b-cb0529192f29.png)

In my Android app my local store is defined as

    Map<String, ColumnDataType> instaTableDefinition = new HashMap<>();
        instaTableDefinition.put("__deleted", ColumnDataType.Boolean);
        instaTableDefinition.put("id", ColumnDataType.String);
        instaTableDefinition.put("inspecteddatetime", ColumnDataType.Date);
        instaTableDefinition.put("inventory", ColumnDataType.String);
        instaTableDefinition.put("wall", ColumnDataType.String);
        instaTableDefinition.put("floor", ColumnDataType.String);
        instaTableDefinition.put("ceiling", ColumnDataType.String);
        instaTableDefinition.put("inspected", ColumnDataType.Boolean);
        instaTableDefinition.put("rejected", ColumnDataType.Boolean);
        instaTableDefinition.put("inspectionid", ColumnDataType.String);
        instaTableDefinition.put("__version", ColumnDataType.String);
        instaTableDefinition.put("levelsininspectionid", ColumnDataType.String);

Sometimes however (not always), the app will run into the exception below, stating that there is no column named ``inspection`` in my table. The Inspection field is a virtual field from the backend and is not supposed to be returned or contain data. If I add the column ``inspection`` to my local store the error will disappear but nothing will be written in the column. It seems like the local sync expects that everything from the server is present in the local DB no matter what type it is. In this case the error can be "fixed" by adding the field, but as stated in issue #534 this error is also present in tables that are provided by the SDK and where the code is not modifiable.

[...]

Issue #505, opened 2014-11-21
[Android] Android SDK throws an exception if the client and server table columns do not match


soninaren:
If an android client queries table data from the server and the server has additional columns like "__deleted" or a user column the Android SDK is throwing an exception. This is not the expected behavior. 

Exception:
com.microsoft.windowsazure.mobileservices.table.sync.localstore.MobileServiceLocalStoreException: android.database.sqlite.SQLiteException: table todoitem has no column named __deleted (code 1): , while compiling: INSERT OR REPLACE INTO "todoitem" ("id","__version","__deleted","complete","text") VALUES (@p1,@p2,@p3,@p4,@p5);

PR #557, opened 2015-01-15
Fix Issues #534 #536 #505 

marianosz:
Fix issues:
SQLite error in table __errors in Android 2.0.1-beta #534
Android SDK thinks virtual fields from backend are missing columns #536
[Android] Android SDK throws an exception if the client and server table columns do not match #505


sdk/android/src/sdk/src/main/java/com/microsoft/windowsazure/mobileservices/table/sync/localstore/MobileServiceLocalStore.java
```diff
@@ -71,18 +71,20 @@
      *
      * @param tableName the local table name
      * @param item      the item to be inserted
+     * @param fromServer
```
aziel: For completeness can you add the doc for the 'fromServer' parameter?  In managed we have:
"-true- if the call is made based on data coming from the server e.g. in a pull operation; -false- if the call is made by the client, such as insert or update calls on an IMobileServiceSyncTable.
"

PR #558, opened 2015-01-16
Add comments for fromServer, taken from Managed client.

aziel:
added comments for fromServer parameter in upsert methods.


aziel:
Mariano took this in his pull request.

PR #562, opened 2015-01-18
[iOS] Adjust bools to be sent as true/false

phvannor:
When using the query params dictionary, boolean values where previous
sent as 1 or 0, now they will be sent as true or false in the query
string.

Updated tests on SyncTable to pass on both 32 and 64 bit systems, since the old method was making some incorrect assumptions on ordering of querystring params.


phvannor:
Will use simpler fix for beta to ship quickly, look at how to handle bools in params later on

PR #563, opened 2015-01-19
[iOS] Update pull to send include deleted flag as true (string) instead ...

phvannor:
...of boolean

### DIFF
--- a/sdk/iOS/src/MSSyncContext.m
+++ b/sdk/iOS/src/MSSyncContext.m
@@ -343,14 +343,8 @@ static NSOperationQueue *pushQueue_;
     }
     
     if (!error && isDeletedParams.count > 0) {
-        // if there are any __includeDeleted params set to NO we want to throw because we would overwrite them
-        for (NSNumber *value in isDeletedParams.allValues) {
-            if (!value.boolValue) {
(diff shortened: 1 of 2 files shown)


sdk/iOS/src/MSSyncContext.m
```diff
@@ -343,14 +343,8 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
     if (!error && isDeletedParams.count > 0) {
-        // if there are any __includeDeleted params set to NO we want to throw because we would overwrite them
-        for (NSNumber *value in isDeletedParams.allValues) {
-            if (!value.boolValue) {
-                error = [self errorWithDescription:@"The '__includeDeleted' parameter value must be YES if used for pullWithQuery:"
-                                      andErrorCode:MSInvalidParameter];
-                break;
-            }
-        }
+        error = [self errorWithDescription:@"The '__includeDeleted' parameter is always true in pullWithQuery: and it's value may not be overridden."
```
brettsam: Nit: Use "its" rather than "it's"

brettsam:
Looks good after that small error message change.

PR #566, opened 2015-01-22
[iOS] update tests to expect encoded comma in urls

phvannor:


phvannor:
Ok, I updated some of the unit tests this would break as another PR.

phvannor:
Goes with PR to encode , to %2C

Issue #559, opened 2015-01-16
java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String com.microsoft.windowsazure.mobileservices.ServiceFilterResponse.getContent()' on a null object reference


pzanzane:
#Android
In class MobileServiceClient.java while using invokeApi methode, app crashes at line given below,
String content = response.getContent();
 response object is null and it is not handled.

invokeApi(apiName, content, httpMethod, requestHeaders, parameters, new ServiceFilterResponseCallback() {
@Override
public void onResponse(ServiceFilterResponse response, Exception exception) {
   if (callback != null) {
   if (exception == null) {
   String content = response.getContent();
   JsonElement json = new JsonParser().parse(content);
   callback.onCompleted(json, null, response);
   } else {
   callback.onCompleted(null, exception, response);
   }
   }
}
});

PR #586, opened 2015-02-11
Fix Issue #559

marianosz:
Fix Issue #559 and Unit test


pzanzane:
Line:1210
What if "response" object is null
String content = response.getContent();

Issue #612, opened 2015-03-09
Executing local operation on MSSyncTable during a push operation prevents storage of the server version after completing the operation


baumatron:
Hey folks,

First I wanted to say thanks for building a cool product that has made building a cloud based app much easier than it could have been. I've been impressed with the offerings of mobile services and overall it seems like a great piece of infrastructure.

I have encountered a few issues as I've been developing an iOS app using over the past few months, and seeing that other community members have posted issues here, I thought I'd try to either bring to light issues that need to be resolved and try to suggest solutions, or have someone correct me and explain to me how a client should actually be interacting with the SDK.

The first issue I'd like to discuss (I'll open issues in the future for the other items) has to do with offline sync and a single client updating and pushing changes to the service. Here's the scenario:

Let's say the app and the service start in a state which is in sync. They both have a single item with a message property equal to "Eat tacos" and a hypothetical version equal to 1. There is only a single client interacting with the server through an MSSyncTable. The following events occur:

1. An update on MSSyncTable changes message -> "Eat pizza", which queues a pending remote update table operation.
2. A push is invoked on the sync context.
3. While the push is in progress (the update operation is locked), another local update is performed on the table changing the message to -> "Eat ice cream", which also enqueues another remote update operation since the previous update operation for that item is locked (it isn't "condensed").
4. The first push update operation completes successfully without conflicts (no server modifications had been applied since the last push), however because there are still pending operations, the result from the server is not stored in the local database, including the new server version.
5. The next push update operation is processed, which pushes the "Eat ice cream" version of the item, which is in fact the subsequent version based on the "Eat pizza" item, but a sync conflict is generated because the server version on the item is stale.

I would expect that in the case of there being a pending operation when step 4 above completes that the server version is saved to the local database, which would allow the subsequent update to complete without conflict.

Thoughts on this?

Thanks very much for your help!

Issue #493, opened 2014-11-14
Cancel pending operations


jcmm33:
Is there no way to cancel pending operations other than spinning up a new database?

We have the requirement for the mobile user to sometimes say 'just refresh from the server and loose anything I've done locally'.  Is the only way of doing this is spin up a new database ?

phvannor:
Following commit in the 1.3.0 release: https://github.com/Azure/azure-mobile-services/commit/aa51d9fae89d734e8ef515501c37a79c67ef81a9
Allows purge to abort all pending operations on the table.

patsissons:
I noticed today that I am able to purge the pending changes, but it won't let me JUST purge the pending changes and not the actual table data. i.e.,
```
// this will delete pending changes, and then purge all rows in table
table.PurgeAsync("queryid", null, true, cancel).Wait();

// this will run if there are no pending changes
// if there are pending changes this throws an InvalidOperationException:
// "The table cannot be purged because it has pending operations."
table.PurgeAsync("queryid", "$filter=false", true, cancel).Wait();
```
I'm not sure why you can't force purge with a filter, anyone care to explain?

phvannor:
The general idea, to update purge to run the filter to get record Ids, purging just those pending ops, then purging the associated table data is possible.  The main complication would be how to deal with deletes in the queue. Since those are not in the SQL db, we'd need to figure out if we try to run the query against the stored items in the pending ops queue or ignore them and leave them pending.

At this time, it was mainly left out for simplicity reasons and to wait to see if it was a requested feature.

patsissons:
what do you recommend to simply wipe out any pending changes? I can manually scan the operations table to see which entities should be deleted and delete them before finally clearing the operations table (and errors table too?). I'm not sure if that is the best approach though, I was mostly hoping that your purge pending changes code would be available to call directly.

phvannor:
Ideally you could call the syncContext CancelAndUpdateItemAsync or CancelAndDiscardItemAsync for each op.  It looks like for some reason that takes an error and not an operation... Those two give you control over what you want to end up in the local db, while (hopefully) hiding implementation details.  Sadly you need to make an op error and probably get the op by doing a manual read.  Something we should improve here to make it easier to scan your ops, and handle them w/o triggering a push in this case.

patsissons:
[...]

Issue #413, opened 2014-09-23
Tracking operation state i.e. pending/attempted and throwing exception on collapse if insert was attempted and delete tries to collapse it (To prevent orphan on server)


yavorg:
If an insert has been attempted but timed out or network connectivity error caused us to lose the response then delete can collapse and cancel the insert while it may have already been completed on server leaving an untracked orphan on the client. To avoid such a problem we need to track operation state i.e. pending, attempted, done. See 10a362eb for details.

ysxu:
@marianosz investigated and this does not apply to android sdk

phvannor:
Can you provide details on how this does not apply?

phvannor:
Also "while it may have already been completed on server leaving an untracked orphan on the client", is a little unclear.  It should read, "while it may have already been completed on server, losing track of the user's delete request. The server will keep that item, and the client state will be incorrect until the next time the item is modified."

ysxu:
@marianosz can you follow up on this?

marianosz:
Our implementation keeps all the offline operations on a queue.
When the push comes, the operations are iterated, and one by one send to the server, and after that removed from the queue, no matter that exists an error or not.
If the insert operation is waiting to be processed and a delete operation appears, the insert operation will be collapsed and both operation will be discarted.
If the insert operation begin to be processed, it will be locked and there is no way to be collapsed.
Summarizing, there is no way to the delete operation to collapse on a failed insert operation, so we are never going to find orphan server elements on this scenario.

masojus:
I'm not sure why the Android tag was added here, but the commit referenced is all related to the Managed SDK, so I'll chime in here to say that I've seen the .NET SDK attempt to collapse operations onto stale InsertOperations, regardless of success or not, due to a few race conditions that exist. Also, in common cases we can see operations not get removed from the OperationQueue because their operation version is bumped on another thread, or stale operations are shoved back into the queue accidentally during an operation update that happens due to a race condition. In short, the scenario described in this issue is at least a subset of issue #615 which I opened to talk about behavior I demonstrate in the unit test in pull request #614. Please take a look at that issue and pull request for more information.  Thanks!

phvannor:
@masojus This issue is tracking a bug fix made to the managed SDK that needed to be checked if  Android had the same bug.  Feel free to add any additional context to #615 as needed then so it can be looked at for the managed code base.

phvannor:
[...]

PR #614, opened 2015-03-09
Add a unit test to exhibit bad behavior when creating pending operations...

masojus:
... during an in-flight PushAction.

Please see issue #615 for more information.

The core problem is synchronization between pending operations and the PushAction. We end up with stale versions, stale operations in the OperationQueue, erroneously-cancelled pending operations, false merge conflicts, inconsistency between local store and back-end, SyncErrors being swallowed and/or exceptions thrown for a variety of reasons.

The setup is that we install a SyncHandler that will do an async wait before executing its operation, and the Unit Test will do the following:

1) Insert an item via the sync table.
2) PushAsync()
3) UpdateAsync() the same item.
4) UpdateAsync() the same item again.
5) Release() the semaphore the SyncHandler is waiting on, which allows the PushAsync() call to continue.
6) await the push
7) await the first update
8) await the second update
9) Push again.
10) Release the push again.

The expectation is that the InsertOperation is executed on the back-end during the first Push, and the two UpdateOperations are collapsed into one, which is effected on the back-end during the second push. This isn't what happens, however.

In the Issue and linked txt document, I have a much more detailed description of the bugs I found in this scenario with call stacks and play-by-play. I've hit this scenario very frequently in a real app with normal user input, but the Unit Test exhibits the behavior very consistently.

A few related issues are #413, #493 and #612 .

A possible solution for part of the problem is to lift the LockItemAsync() call out of ExecuteOperationAsync() out into ExecuteAllOperationsAsync() to make the calls to ExecuteOperationAsync(), OperationQueue.DeleteAsync(), and OperationQueue.PeekAsync() atomic, but I haven't tested this yet.

Thank you!


azurecla:
Hi __@masojus__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@masojus__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -370,6 +372,17 @@ private Task ExecuteOperationAsync(MobileServiceTableOperation operation, JObjec
 
                 try
                 {
+                    if (MobileServiceTableOperationKind.Insert != operation.Kind)
+                    {
+                        // Refresh the item's version, in case it changed while we were awaiting the item lock
+                        var localStoreItem = await this.Store.LookupAsync(operation.TableName, operation.ItemId);
```
phvannor: This is essentially enforcing a last write wins policy, we can't assume the pending item on the client can just overwrite the server's item without a conflict.  The server could easily have changed additional fields/etc that the prev client object may now accidently stomp on.

I can see the usefulness of this option, based on the specific app/table, but not generally.

masojus: Correct. To be clear, I don't think this change should be taken as-is, but rather is just to show what I'm talking about in the comments for the previous commit and for issue #615. In the comments for this change, I state that "it doesn't account for when the initial push generates a legitimate merge conflict on the back-end, and the resolution is to accept the server version," which is exactly what you're saying. I think there has to be more infrastructure to call into client code and ask for the intent.  In my own client code, I call back into an IConflictResolver and get back a tuple with a new type-specific item plus an enum value that is either KeepLocal, KeepServer or ManualMerge, and the code in this change is indeed valid for the KeepLocal case.  In my own code, I have to fudge this in client code, which rapidly gets super complicated, and in the KeepServer version, to fix up the operations queue, I have to go manually delete rows from the operations table, which seems sketchy.  Thoughts?

phvannor: Yes, I completely agree and we need a more flushed out concept of client conflicts for these scenarios. 

I think the delegate needs some new methods on it, like a updateReceivedForItem:, and let the client code respond & update their local copy as needed.  For operations that were blocked, they would also need to trigger something similar once they get the lock again and know how to resolve the "local conflict".  

I'll need to review my earlier notes & discuss internally on what our plans are on this area again.  Please continue to let us know the pain points you are seeing so we can prioritize correctly.

masojus: Did you get a chance to look at the document I linked to in issue #615? It's much more descriptive and complete in terms of the various problems that occur due to the race condition.

Regarding the versioning issue, now that I re-read your initial comment, are you suggesting that my fix could cause us to stomp on server changes in the single-client case? I realize it's bad in the case where the first push causes an actual conflict, where the server really has a future version from another client updating it, but I'm not sure I see how we could trample changes to other columns made by the operation(s) in the first push, because the change only affects what is written locally when the operation gets a chance to execute and first wants to effect the change to the local store.  Later, the item is pulled from the store in the PushAction by calling LoadOperationItem(), and there it will have all the columns as they were when we returned from the last remote execution in the previous push, but with the last __version we know about from the server.  If no other client changes that row, this seems like the correct behavior.  Only once the row in the local store has been tainted with server data (from a pull or a merge conflict) is this incorrect, and in that case all pending operations have to be invalidating.

I'm investigating a workaround for now for the later case: I plan to keep all my conflict resolution code, which is pretty complicated already without these issues that cropped up, but in the case where I resolve a conflict by picking the server item, I was hoping to maybe use the CancelAndDiscard*() API, except that is only available in OnPushComplete(), so somehow I'd have to know that in the SyncHandler I need to fail all subsequent operations *for this push*, but I don't think we have the context in SyncHandler to know if we're in one push or another...plus due to the race this change fixes, you can keep adding operations to the OperationQueue even while a push is in-flight.

I'm curious: is this stuff being actively discussed internally? Are these known issues you're working on, or is this back-burner stuff? I'm wondering how much effort to put into working around these issues in client code, and/or how much time to devote to fixing it in my fork of the SDK, or if something is in the pipeline to be fixed soonish.  Thanks!

phvannor:
Reviewing #615 as well, the place to start is moving the delete before releasing the lock, which it looks like you did here.  If you want to submit that portion against Dev we could merge it in.

For my point on item, when it runs item will be saved into the localstore, so any column present in item will be overwritten.  So it really depends on what gets stomped based on the actual client code in use.  (and also the server logic, does it do any simple merges, etc)  

@lindydonna is currently working on our next set of offline features.  If its easier you can email mobileservices@microsoft.com, with contact details and we can look at setting up a call to go over your experiences/issues/etc and what you would like to see going forward.

masojus:
This was handled by Pull Request #633 into 'dev' branch, so I'm closing this one.

PR #617, opened 2015-03-09
Fix Issue 613

marianosz:
Fix Issue 613

### DIFF
--- a/sdk/android/src/sdk/src/main/java/com/microsoft/windowsazure/mobileservices/table/query/QueryNodeODataWriter.java
+++ b/sdk/android/src/sdk/src/main/java/com/microsoft/windowsazure/mobileservices/table/query/QueryNodeODataWriter.java
@@ -2,19 +2,19 @@
 Copyright (c) Microsoft Open Technologies, Inc.
 All Rights Reserved
 Apache 2.0 License
- 
+
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
- 
+
      http://www.apache.org/licenses/LICENSE-2.0
- 
+
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
- 
+
 See the Apache Version 2.0 License for specific language governing permissions and limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


sdk/android/src/sdk/src/main/java/com/microsoft/windowsazure/mobileservices/table/query/QueryNodeODataWriter.java
```diff
@@ -47,7 +47,7 @@ private static String process(String s) {
     }
 
     private static String process(Date date) {
-        return "datetime'" + DateSerializer.serialize(date) + "'";
+        return "datetimeoffset'" + DateSerializer.serialize(date) + "'";
```
phvannor: Is this limited to the updatedAt column only?  We do not want all datetimes to go as datetimeoffset (That would break all Node backends)  This needs to be  al little more involved.

marianosz:
All the test against the node backend are green using datetimeoffset instead of datetime. 
Considering that results, my conclusion is that change the data type to datetimeoffset on the uri is the best fix for this issue and doesn't affect the use of the Node backend.
But I'm open to other suggestions. Thanks

phvannor:
The node runtime will attempt to convert datetimeoffsets back into DateTimes, but it's not intended to be a long term fix.  I thought that fix is limited to appying only to the __updatedAt/createdAt columns, but it may be more general, not 100% sure on that.

Also, this solution will break against .NET backend as well.  Sending datetime vs datetimeoffset depends on the SQL type of the column itself.  The default date column we use for updatedAt should use datetimeoffset.  However, the column type the customer uses could demand datetime just as easily.

As per the managed SDK's solution, we need to have a way for the query generator to determine if it should use datetimeoffset vs datetime.  (For managed that was easy, for iOS we made a new class MSDateOffset the customer uses)

Also, I am unsure why you didn't have to fix any unit test cases.  We should have test cases confirming dates are sent correctly in the QS.  I'd have expected the MobileServiceQueryTests, MobileServiceSyncTableTests, and a couple other files that have hardcoded checks on datetime'x' to also need modified for this.

nikatlas:
My comment here. 
As i saw the code a test like this would clarify things here. 
I think the above solution will arise more problems. ( though i have it on my project )
```javascript
table.where(function (at) {
                            return this.dateField > new Date(at);
                        }, dateAsString)
                .read()
                .done(function (results) {
                          //Success
                }, function (err) {
                          //Failure
                });
```
where dateField is a DateTime on the backend and NOT a DateTimeOffset

ysxu:
fixed with https://github.com/Azure/azure-mobile-services/pull/645

PR #627, opened 2015-03-11
Unit test and potential fix for #612.

baumatron:
The issue here is that when performing sequential updates or an insert followed by an update on the server, the server barfs on the second update because the version provided by the server after the first operation isn't saved and therefore isn't sent to the server when attempting the second operation. See #612 for more details.

This change adds a unit test for this case and a potential fix. All unit tests passing with this change.

### DIFF
--- a/sdk/iOS/src/MSQueuePushOperation.m
+++ b/sdk/iOS/src/MSQueuePushOperation.m
@@ -225,6 +225,43 @@
                                               internalError:storeError];
                 }
             }
+            else {
+                // Operation was successful, but another operation is pending.
+                // To be able to complete the subsequent operation, the version
+                // field needs to be updated to the latest version we just got
+                // from the server.
+                NSError* storeError;
+                
+                // Get the item from the local store, which contains updates that haven't yet been pushed to the server.
+                NSDictionary* storeItem = [self.syncContext.dataSource readTable:operation.tableName
+                                                                      withItemId:operation.itemId
+                                                                         orError:&storeError];
+                if (storeError) {
+                    NSString *errorMessage = [NSString stringWithFormat:@"Unable to read item '%@' from table '%@'",
+                                              operation.itemId, operation.tableName];
+                    
+                    self.error = [self errorWithDescription:errorMessage
+                                                       code:MSPushAbortedDataSource
+                                              internalError:storeError];
+                }
+                else {
+                    // Update the version of the item to the lates version we've seen from the server
+                    NSString* newVersion = [item objectForKey:MSSystemColumnVersion];
+                    [storeItem setValue:newVersion forKey:MSSystemColumnVersion];
+                    
+                    [self.syncContext.dataSource upsertItems:[NSArray arrayWithObject:storeItem]
+                                                                                table:operation.tableName
+                                                                              orError:&storeError];
+                    if (storeError) {
+                        NSString *errorMessage = [NSString stringWithFormat:@"Unable to upsert item '%@' into table '%@'",
+                                                  operation.itemId, operation.tableName];
+                        
+                        self.error = [self errorWithDescription:errorMessage
+                                                           code:MSPushAbortedDataSource
+                                                  internalError:storeError];
+                    }
+                }
+            }
         }
         
         if (self.error) {
--- a/sdk/iOS/test/MSSyncTableTests.m
+++ b/sdk/iOS/test/MSSyncTableTests.m
@@ -19,8 +19,71 @@
 #import "MSSyncContextInternal.h"
 #import "MSTableConfigValue.h"
 
+static NSString *const TodoTable = @"TodoItem";
 static NSString *const TodoTableNoVersion = @"TodoNoVersion";
 static NSString *const AllColumnTypesTable = @"ColumnTypes";
+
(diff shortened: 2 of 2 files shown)


azurecla:
Hi __@baumatron__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@baumatron__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

sdk/iOS/src/MSQueuePushOperation.m
```diff
@@ -225,6 +225,43 @@ - (void) finishTableOperation:(MSTableOperation *)operation item:(NSDictionary *
                                               internalError:storeError];
                 }
             }
+            else {
+                // Operation was successful, but another operation is pending.
```
phvannor: I think you need to compare operation.item with item (minus system properties) for this to be ok to do, as we don't want to stomp on other server changes.

sdk/iOS/src/MSQueuePushOperation.m
```diff
@@ -225,6 +225,43 @@ - (void) finishTableOperation:(MSTableOperation *)operation item:(NSDictionary *
...
+                // To be able to complete the subsequent operation, the version
+                // field needs to be updated to the latest version we just got
+                // from the server.
+                NSError* storeError;
+                
+                // Get the item from the local store, which contains updates that haven't yet been pushed to the server.
+                NSDictionary* storeItem = [self.syncContext.dataSource readTable:operation.tableName
```
phvannor: In the case where the pending operation is a delete, there will be no storeItem.  What you actually want to update is the item in the operation queue.  

You aren't explicitly handling a storeItem = nil case in the below logic.

sdk/iOS/src/MSQueuePushOperation.m
```diff
@@ -225,6 +225,43 @@ - (void) finishTableOperation:(MSTableOperation *)operation item:(NSDictionary *
...
+                    self.error = [self errorWithDescription:errorMessage
+                                                       code:MSPushAbortedDataSource
+                                              internalError:storeError];
+                }
+                else {
+                    // Update the version of the item to the lates version we've seen from the server
+                    NSString* newVersion = [item objectForKey:MSSystemColumnVersion];
+                    [storeItem setValue:newVersion forKey:MSSystemColumnVersion];
+                    
+                    [self.syncContext.dataSource upsertItems:[NSArray arrayWithObject:storeItem]
```
phvannor: Continuing the above, in the pending delete case, this will cause errors.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -340,6 +403,146 @@ -(void) testInsertPushInsertPush
     
 }
 
+
+
+-(void) testInsertUpdateWhilePushing
```
phvannor: Ideally we'd end up with 2 unit tests: update with server changes (leaves version alone), this test case (Server item == sent copy), and a new one for pending delete gets version updated (if server == sent).

phvannor:
Hey Brian, can you let me know if you pla to address my comments or not?  If not, I may take some time to look at adding the approach in myself over the next week or two.

sdk/iOS/src/MSQueuePushOperation.m
baumatron: This would seem to be the place to hook in a local change conflict delegate to resolve any conflicts and allow for preserving server side changes. What do you think about adding something like that here?

sdk/iOS/src/MSQueuePushOperation.m
baumatron: Makes sense.

baumatron:
Sure, I can update the pull request addressing your feedback. Let me know what you think about the local merge conflict handler and I can throw something together for that.

sdk/iOS/src/MSQueuePushOperation.m
phvannor: This is a logical place to handle a conflicts during a push with a saved op.  That doesn't help anything already in memory in the app. 

I'd like to see the logic I outlined as the fallback case I think either way.

If adding a conflict handling delegate method it would ideally be usable from a SyncTable action, and easily called from the tableOperation: onComplete: method as well as here.  Also, potentially from a item change notification during Pulls & Pushes.

phvannor:
I think keeping the local conflict handler as a separate PR may be cleaner so we can take the changes quicker.  I imagine agreeing on conflict handling may be more involved, esp around figuring out how the work will apply to the other SDKs.

paulbatum:
@phvannor this PR is pretty old, where did we end up here? Should we close?

phvannor:
we will have to close this one, but I wanted some of the core concepts to be carried over, but part of it is tied up in a pull handler and/or hooks for post pull/push to resolve unecessary conflicts down stream.  With Naren's work on hold, I guess we hold for reference, and possibly refactor to take a couple of the smaller ideas in when time permits.

adrianhall:
Closing this PR as the Azure Mobile Services repo is frozen in preference of the Azure Mobile Apps work.

Issue #603, opened 2015-02-27
Update SDK with appropriate headers for data on calls against mobile services runtime


ysxu:
We need this for the android SDK usage data.

Zumo runtime will be able to pick up specific headers to determine the calls' SDK version number as well as their used features (offline, etc) for us to better construct a usage story. See the iOS/managed SDK for more details on the header. @phvannor worked with them in case there are any questions.

Thanks,
Mimi

ysxu:
@marianosz 
@soninaren merged the changes in but it seems like the header version still says 2.0.2 beta -- we would want it to be 2.0.2 instead since we are going GA next. Thanks!

PR #629, opened 2015-03-11
Fix UT Issue #603

marianosz:
Fixing of Unit tests for issue #603


phvannor:
Going forward, we are using the commit titles to generate the change log using something like the following command:

git log previousrelease..HEAD --pretty=format:'%s [%h](https://github.com/Azure/azure-mobile-services/commit/%h)' | sls "merge pull|merge branch" –NotMatch

Can you try to make sure Android commits look nice when output that way.

Issue #615, opened 2015-03-09
Creating pending operations during in-flight PushAction has race conditions that cause several bugs


masojus:
Thanks all for the good work on AMS and for providing a lot of functionality that's easy to get working quickly. I wanted to get some feedback on some naughty behavior I'm seeing around Offline Sync when pending operations on the local store are created while a push is currently executing.

See #614 for a Unit Test that exhibits parts of this behavior.
Related issues: #413, #493, #612 

The short version is that if you generate pending operations while a push is in-flight, there are various race conditions that occur, which generate a plethora of bugs. We end up with stale versions, stale operations in the OperationQueue, erroneously-cancelled pending operations, false merge conflicts, inconsistency between local store and back-end, SyncErrors being swallowed and/or exceptions thrown for a variety of reasons.

The reasons are myriad, but fundamentally, the PushAction can fail to delete operations before subsequent operations look at the queue and try to Collapse(), and even if it does delete an operation, we can Enqueue() to the OpQueue at any time, even with an in-flight PushAction, and we Collapse() onto/Enqueue()/Update()/generally use operations that are stale without verification and without any synchronization. This is contrary to the comments in issue #413 indicating that we can't cause an operation to try to Collapse() onto a failed InsertOperation, because my unit test exhibits exactly that behavior. In the best case here, we see the behavior in Issue #612, where we operate with old version numbers for subsequent operations, because they were already waiting on the item lock, with the item version already captured, when the push may or may not have updated the version in the local store, so all waiting operations will cause a merge conflict with the back-end. In the worst case, we throw, or we cancel pending operations, delete SyncErrors from the errors list without notifying client code, end up with inconsistencies between the back-end and the local store, and put stale operations back into the OpQueue.

Please take a look at [this document](https://onedrive.live.com/?cid=3E227B64092E6F3B&id=3E227B64092E6F3B%21114133) for a full description of the 12-14 bugs that I found this afternoon while looking a little deeper into this odd behavior. Note that the document has call stacks in it, but if you scan down, you'll see the text "Bug #" along with a description of each bug as it occurs while walking through the scenario the unit test in pull request #614 encapsulates.

Thanks in advance for any feedback and/or advice!

masojus:
[...]

PR #633, opened 2015-03-12
Fix a large part of the problem described in Issue #615. This is a subse...

masojus:
...t of Pull Request #614 with just the subset I think can be merged as per the conversation with @phvannor.

This moves the lock out to protect the calls to DeleteAsync() and PeekAsync() on the OperationQueue, which prevents a lot of the badness described in Issue #615. There are still deeper issues around versioning of the actual items in the local store, but that's another discussion. All unit tests pass on WP8, WP81 and Win8. 

I'm requesting this be pulled into the 'dev' branch as per the comment from @phvannor on Pull Request #614.

### DIFF
--- a/sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/Actions/PushAction.cs
+++ b/sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/Actions/PushAction.cs
@@ -132,104 +132,104 @@ namespace Microsoft.WindowsAzure.MobileServices.Sync
             // keep taking out operations and executing them until queue is empty or operation finds the bookmark or batch is aborted 
             while (operation != null)
             {
-                bool success = await this.ExecuteOperationAsync(operation, batch);
-
-                if (batch.AbortReason.HasValue)
+                using (await this.OperationQueue.LockItemAsync(operation.ItemId, this.CancellationToken))
                 {
-                    break;
-                }
+                    bool success = await this.ExecuteOperationAsync(operation, batch);
 
-                if (success)
-                {
-                    // we successfuly executed an operation so remove it from queue
-                    await this.OperationQueue.DeleteAsync(operation.Id, operation.Version);
-                }
+                    if (batch.AbortReason.HasValue)
+                    {
+                        break;
+                    }
 
-                // get next operation
-                operation = await this.OperationQueue.PeekAsync(operation.Sequence, this.tableKind, this.tableNames);
+                    if (success)
+                    {
+                        // we successfuly executed an operation so remove it from queue
+                        await this.OperationQueue.DeleteAsync(operation.Id, operation.Version);
+                    }
+
+                    // get next operation
+                    operation = await this.OperationQueue.PeekAsync(operation.Sequence, this.tableKind, this.tableNames);
+                }
             }
         }
 
         private async Task<bool> ExecuteOperationAsync(MobileServiceTableOperation operation, OperationBatch batch)
         {
-            using (await this.OperationQueue.LockItemAsync(operation.ItemId, this.CancellationToken))
+            if (operation.IsCancelled || this.CancellationToken.IsCancellationRequested)
             {
-                if (operation.IsCancelled || this.CancellationToken.IsCancellationRequested)
-                {
-                    return false;
-                }
-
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@masojus__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

phvannor:
Code wise this look OK, any chance you can add a unit test that verifies the locking is working as intended?  I believe #633's could be tweaked to just focus on this particular problem.

masojus:
If you mean the Unit Test in #614, then yes, I can remove the part about validating version numbers, since that issue still remains, and just force the situation that can cause the OperationQueue corruption and make sure that's fixed.

phvannor:
Oops yes, I meant #614.

masojus:
Commit 5c16e37 address @phvannor's request for a Unit Test. This doesn't test anything regarding versioning issues around pending operations and such, but focuses on the race condition around changes made to the OperationQueue.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/Actions/PushAction.cs
```diff
@@ -132,104 +132,104 @@ internal class PushAction : SyncAction
...
-                // get next operation
-                operation = await this.OperationQueue.PeekAsync(operation.Sequence, this.tableKind, this.tableNames);
+                    if (success)
+                    {
+                        // we successfuly executed an operation so remove it from queue
+                        await this.OperationQueue.DeleteAsync(operation.Id, operation.Version);
+                    }
+
+                    // get next operation
+                    operation = await this.OperationQueue.PeekAsync(operation.Sequence, this.tableKind, this.tableNames);
```
hasankhan: this doesn't have to be inside the lock. Only execution and removal of operation from queue has to be synchronized, not peeking the next operation.

sdk/Managed/test/Microsoft.WindowsAzure.MobileServices.Test/UnitTests/Table/Sync/MobileServiceSyncContext.Test.cs
```diff
@@ -246,6 +246,136 @@ public class MobileServiceSyncContextTests : TestBase
...
+            /// </summary>
+            /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
+            internal void SetLookupDelay(int delayInMilliseconds)
+            {
+                _lookupDelayInMilliseconds = delayInMilliseconds;
+            }
+        }
+
+        [AsyncTestMethod]
+        public async Task PushAsync_Succeeds_WithPendingOperations_AndOpQueueIsConsistent()
```
hasankhan: A simpler test would be to block the deletion of operation from the store and attempt an operation on the same item. If the operation succeeds then we have an error, otherwise test passes. You can put a timeout on second operation (because it won't get access to lock as the delete of operation is blocked).

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/Actions/PushAction.cs
masojus: This has already been merged, but in any case, it's not absolutely necessary that the peek be under the lock, but it's a step in the right direction, as this "batch" isn't really a batch at all since operations keep creeping into the push action while in-flight.

sdk/Managed/test/Microsoft.WindowsAzure.MobileServices.Test/UnitTests/Table/Sync/MobileServiceSyncContext.Test.cs
masojus: Blocking on DeleteAsync() in the store gives OperationQueue.DeleteAsync() a chance to call GetOperationAsync() first. To ensure the conditions necessary to prove what this test proves, it's necessary to inject the delay as early as possible in OperationQueue.DeleteAsync(), and the first thing that happens, effectively, is the LookupAsync() call on the local store.

Issue #568, opened 2015-01-25
[offline][iOS] Item already inserted doesn't create operation message


damienpontifex:
I have previously been using the iOS offline SDK. Although I use actual NSManagedObject instances in my view controllers and override the `-[NSManagedObject willSave]` method which I then do the appropriate insert/update/delete on the syncTable with entityName.

This being the case, the new code in `MSSyncContext` at https://github.com/Azure/azure-mobile-services/blob/offline/sdk/iOS/src/MSSyncContext.m#L160-L176 throws an error when an insert occurs but the item already exists in the local store.

Could these two case statements be merged seeing as the actual operation is just an upsert? i.e.

```objective-c
case MSTableOperationUpdate:
case MSTableOperationInsert:
                    [self.dataSource upsertItems:[NSArray arrayWithObject:itemToSave] table:table orError:&error];
                    break;
```

And then the data source will continue to do the appropriate insert into the operation table for the server push to occur.

phvannor:
Throwing an error when inserting an existing item that already was present locally, was a purposeful bug fix.  I see how this breaks your usage pattern though.  

If the insert read set the includesPendingChanges to NO in its fetch request, that would also fix this for you correct?  (Although, that probably requires a new method in MSSyncContextDataSource)

You are using the same MOC in your ViewControllers and the MSSyncConext?

damienpontifex:
Just tried adding that to the fetch request in `-[MSCoreDataStore getRecordForTable:itemId:asDictionary:orError:]` but it still returned the object in the results. Hence, same problem for me.

For reference, this insert is called in my custom subclass will save method but before `[super willSave]` so not sure why this object still wouldn't be pending as I haven't completed the will save method flow.

Also, yes the MSCoreDataStore is instantiated with the managedObjectContext property of the saving object, so same context.

damienpontifex:
@phvannor is there any updated guide for using NSManagedObject instances in code rather than the NSDictionary approach? I was trying to attain automatic saving to server messages. As a save to the context can cause a save of all objects in memory, sometimes it is troublesome to determine which objects to actually call the save method with the MSCoreDataStore to create the message.

I had also previously raised #287 but can see problems now with my implementation this case plus purge operations (which with my current setup will actually cause a remote delete, not just a local purge).

damienpontifex:
[...]

PR #635, opened 2015-03-12
[iOS] NSManagedObjectContext save observer

damienpontifex:
Context observer for automatic message creation for entities saved with an NSManagedObjectContext. Also prevents the MSCoreDataStore from trying to re-insert/update/delete the entity and just creates the server message as discussed in issue #568

### DIFF
--- a/sdk/iOS/src/MSManagedObjectObserver.m
+++ b/sdk/iOS/src/MSManagedObjectObserver.m
@@ -7,35 +7,115 @@
 //
 
 #import "MSManagedObjectObserver.h"
+#import "MSCoreDataStore.h"
+#import "MSClient.h"
 #import <CoreData/CoreData.h>
 
 @interface MSManagedObjectObserver()
-@property (nonatomic, weak) MSClient *client;
+@property (nonatomic, strong) MSClient *client;
+@property (nonatomic, weak) NSManagedObjectContext *context;
 @end
 
 @implementation MSManagedObjectObserver
 
-- (id) initWithClient:(MSClient *)client
-{
-    return [self initWithClient:client contextsToObserver:nil];
-}
-
-- (id) initWithClient:(MSClient *)client contextsToObserver:(NSArray *)contexts
+- (instancetype) initWithClient:(MSClient *)client
 {
     self = [super init];
     if (self) {
-        _client = client;
+		// Copy so we can change the handling of how the sync table operations are handled
+        _client = [client copy];
+		
+		if ([_client.syncContext.dataSource isKindOfClass:[MSCoreDataStore class]])
+		{
+			MSCoreDataStore *dataStore = _client.syncContext.dataSource;
+			_context = dataStore.context;
+			
+			[[NSNotificationCenter defaultCenter] addObserver:self
+													 selector:@selector(handleDidSaveNotification:)
+														 name:NSManagedObjectContextDidSaveNotification
+													   object:_context];
+		}
+		
+		// Modify the handling of sync table operations for this instance of the data source
+		_client.syncContext.dataSource.handlesSyncTableOperations = FALSE;
     }
     return self;
 }
 
-- (void) registerAsObserver
+- (void)dealloc
 {
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(handleDidSaveNotification:)
-                                                 name:NSManagedObjectContextDidSaveNotification
-                                               object:nil];
+	if (self.context != nil)
+	{
+		[[NSNotificationCenter defaultCenter] removeObserver:self
+														name:NSManagedObjectContextDidSaveNotification
+													  object:self.context];
+	}
+}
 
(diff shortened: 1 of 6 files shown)


azurecla:

TTYL, AZPRBOT;

sdk/iOS/src/MSManagedObjectObserver.m
```diff
@@ -7,35 +7,122 @@
...
+
+/// Checks the entity has the ms_version attribute to signal it should
+/// be synchronised to the mobile service
+- (BOOL)entityHasRemoteStoreAttributes:(NSManagedObject *)managedObject
 {
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(handleDidSaveNotification:)
-                                                 name:NSManagedObjectContextDidSaveNotification
-                                               object:nil];
+	return (managedObject.entity.attributesByName[StoreVersion] != nil);
```
phvannor: I didn't really intend to require all LocalStores to use ms_version long term.  Or actually require the use of version in the table, optimistic concurrency while recommended, is not technically required.

I originally was thinking we'd just ignore "ms_" tables, but I see how this lets the dev use tables for their own purposes as well.  

For prototypes this is fine, long term we may need to revisit this logic.

sdk/iOS/src/MSManagedObjectObserver.m
```diff
@@ -7,35 +7,122 @@
...
+}
 
+- (void)handleDidSaveNotification:(NSNotification *)notification
+{
+	NSSet *insertedObjects = notification.userInfo[NSInsertedObjectsKey];
+	for (NSManagedObject *insertedObject in insertedObjects)
+	{
+		if ([self entityHasRemoteStoreAttributes:insertedObject] == FALSE) {
```
phvannor: == NO?

sdk/iOS/src/MSManagedObjectObserver.m
damienpontifex: I have a couple of tables just for local information. The only other way I saw around this at this stage was having separate stacks for entities to be synced and local. This implementation seemed like the lesser work.

sdk/iOS/src/MSManagedObjectObserver.m
damienpontifex: Sorry, too much swift lately. Also wasn't sure on coding style so I thought more verbose would be the safer bet. Will update shortly

sdk/iOS/src/MSManagedObjectObserver.m
```diff
@@ -7,35 +7,122 @@
...
+			// Only apply table operations on entities we expect to be managed by the mobile service
+			continue;
+		}
+		
+		NSString *tableName = insertedObject.entity.name;
+		NSDictionary *tableItem = [MSCoreDataStore tableItemFromManagedObject:insertedObject];
+		
+		MSSyncTable *syncTable = [self.client syncTableWithName:tableName];
+		[syncTable insert:tableItem completion:^(NSDictionary *item, NSError *error) {
+			if (error != nil) {
```
phvannor: This is interesting, I suppose we expect syncTable failures to only be dev time generally.  I'm trying to think what normal usage would see in an app.  

Should this raise a new notification to the developer to handle?  Right now this just leaves the app in a bad state.

phvannor: Inserts, could in theory fall back to an updates, depending on the error (id already exists?)

phvannor:
I have some general comments, but I think the core is good.  Since this is a prototype branch, I'm happy taking this as is.  Let me know if you want to address anything right now.

sdk/iOS/src/MSManagedObjectObserver.m
damienpontifex: Another notification, an error handler callback block on the MSManagedObjectObserver instance or delegate protocol?

My general preference would be the callback block as it can be written inline with instantiating the observer.

damienpontifex:
Made the small changes for the boolean check and some comments above on the error handling. Happy to merge and continue to test and work on otherwise

sdk/iOS/src/MSManagedObjectObserver.m
phvannor: A callback block is fine I think.  I need to think a little on what the error(s) are likely to be, and what the typical resolution would be for them.

sdk/iOS/src/MSManagedObjectObserver.m
phvannor: Makes sense. I have a couple designs in flight (like dealing with relationships) that would require re-visiting this anyways later on.  It doesn't seem unreasonable for now to put this restriction on using the observer.

phvannor:
OK, I am going to merge this in, as I would like to try out the changes/etc and I know others have expressed interest in this topic.

damienpontifex:
I have added some unit tests for this class. Should I just create another pull request to include them?

PR #637, opened 2015-03-13
Unit tests for new managed object observer

damienpontifex:
Unit tests to support PR #635 and associated with issue #568

### DIFF
--- a/sdk/iOS/src/MSManagedObjectObserver.h
+++ b/sdk/iOS/src/MSManagedObjectObserver.h
@@ -18,6 +18,10 @@ typedef void (^MSManagedObjectObserverCompletionBlock)(MSTableOperationTypes ope
 - (instancetype) initWithClient:(MSClient *)client;
 
 /// Block to be called on each operation that will is inserted into MS_TableOperations
+///
+/// If not implemented or not dealing with errors and errors occur during these operations
+/// the local store and associated messages to the mobile services API will be left
+/// in a bad state.
 @property (nonatomic, copy) MSManagedObjectObserverCompletionBlock observerActionCompleted;
 
 @end
--- a/sdk/iOS/src/MSManagedObjectObserver.h
+++ b/sdk/iOS/src/MSManagedObjectObserver.h
@@ -7,11 +7,17 @@
 //
 
 #import <Foundation/Foundation.h>
+#import "MSTableOperation.h"
+
+typedef void (^MSManagedObjectObserverCompletionBlock)(MSTableOperationTypes operationType, NSDictionary *item, NSError *error);
 
 @class MSClient;
 
 @interface MSManagedObjectObserver : NSObject
 
 - (instancetype) initWithClient:(MSClient *)client;
+
+/// Block to be called on each operation that will is inserted into MS_TableOperations
+@property (nonatomic, copy) MSManagedObjectObserverCompletionBlock observerActionCompleted;
 
 @end
--- a/sdk/iOS/test/MSManagedObjectObserverTests.m
+++ b/sdk/iOS/test/MSManagedObjectObserverTests.m
@@ -1,0 +1,160 @@
+//
+//  MSManagedObjectObserverTests.m
+//  WindowsAzureMobileServices
+//
+//  Created by Damien Pontifex on 13/03/2015.
+//  Copyright (c) 2015 Windows Azure. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+#import "MSCoreDataStore.h"
+#import "MSCoreDataStore+TestHelper.h"
+#import "MSJSONSerializer.h"
+#import "TodoItem.h"
+#import "MSManagedObjectObserver.h"
+#import "MSTableOperation.h"
+
+@interface MSManagedObjectObserverTests : XCTestCase
+@property (nonatomic, strong) MSCoreDataStore *store;
+@property (nonatomic, strong) NSManagedObjectContext *context;
+@property (nonatomic, strong) MSManagedObjectObserver *observer;
+
+@property (nonatomic, strong) TodoItem *item;
+@end
+
+@implementation MSManagedObjectObserverTests
+
+- (void)setUp
+{
+    [super setUp];
+	
+	self.context = [MSCoreDataStore inMemoryManagedObjectContext];
+	self.store = [[MSCoreDataStore alloc] initWithManagedObjectContext:self.context];
+	
+	MSClient *client = [[MSClient alloc] initWithApplicationURL:nil applicationKey:nil];
+	client.syncContext = [[MSSyncContext alloc] initWithDelegate:nil dataSource:self.store callback:nil];
+	self.observer = [[MSManagedObjectObserver alloc] initWithClient:client];
+	
+	
+	self.item = [NSEntityDescription insertNewObjectForEntityForName:@"TodoItem" inManagedObjectContext:self.context];
+	
+	self.item.text = @"Test item";
+	self.item.id = @"ABC";
+	
+	NSError *err;
+	[self.context save:&err];
+}
+
+- (void)tearDown
+{
+	self.store = nil;
+	
+    [super tearDown];
+}
+
+- (void)testObservingInsertOperation
+{
+	NSError *err;
+	
+	/// The save notification and subsequent calls to create table operations is ansynchronous so we have to wait a bit for this process to occur
+	/// TODO: Improve how this is handled. Maybe a callback block for each operation action for
+	sleep(1);
+	
+	NSFetchRequest *tableOperationsRequest = [NSFetchRequest fetchRequestWithEntityName:@"MS_TableOperations"];
+	NSArray *tableOperations = [self.context executeFetchRequest:tableOperationsRequest error:&err];
+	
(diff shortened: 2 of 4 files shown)


azurecla:

TTYL, AZPRBOT;

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	client.syncContext = [[MSSyncContext alloc] initWithDelegate:nil dataSource:self.store callback:nil];
+	self.observer = [[MSManagedObjectObserver alloc] initWithClient:client];
+	
+	
+	self.item = [NSEntityDescription insertNewObjectForEntityForName:@"TodoItem" inManagedObjectContext:self.context];
+	
+	self.item.text = @"Test item";
+	self.item.id = @"ABC";
+	
+	NSError *err;
```
phvannor: nit: we might as well just pass nil for error since we dont use err

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	
+	/// The save notification and subsequent calls to create table operations is ansynchronous so we have to wait a bit for this process to occur
+	/// TODO: Improve how this is handled. Maybe a callback block for each operation action for
+	sleep(1);
+	
+	NSFetchRequest *tableOperationsRequest = [NSFetchRequest fetchRequestWithEntityName:@"MS_TableOperations"];
+	NSArray *tableOperations = [self.context executeFetchRequest:tableOperationsRequest error:&err];
+	
+	for (NSManagedObject *managedObject in tableOperations)
```
phvannor: wouldn't it be simpler to not have setup insert an item than deleting its pending op?

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	{
+		// Clean out the pending insert operation
+		[self.context deleteObject:managedObject];
+	}
+	[self.context save:&err];
+	
+	self.item.text = @"Test item updated";
+	[self.context save:&err];
+	
+	sleep(1);
```
phvannor: You almost need the observer to have a block to call when it finishes saving.  I suppose you could listen for a save notifications from the context its on?

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	
+	tableOperations = [self.context executeFetchRequest:tableOperationsRequest error:&err];
+	
+	XCTAssertEqual(tableOperations.count, 1, @"Should have one insert operation after the save of TodoItem %@", self.item);
```
phvannor: nit: update operation

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	NSManagedObject *tableOperation = tableOperations.firstObject;
+	NSString *operationTable = [tableOperation valueForKey:@"table"];
+	XCTAssertEqualObjects(operationTable, self.item.entity.name, @"The operation should be associated for the %@ table", self.item.entity.name);
+	
+	NSString *operationItemId = [tableOperation valueForKey:@"itemId"];
+	XCTAssertEqualObjects(operationItemId, self.item.id, @"The operation should be associated for the inserted item with id %@", self.item.id);
+	
+	NSDictionary *properties = [[MSJSONSerializer JSONSerializer] itemFromData:[tableOperation valueForKey:@"properties"] withOriginalItem:nil ensureDictionary:YES orError:&err];
+	
+	XCTAssertEqual([properties[@"type"] integerValue], MSTableOperationUpdate, @"Associated operation should be an insert with newly created object");
```
phvannor: update again

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	NSString *originalItemId = self.item.id;
+	
+	[self.context deleteObject:self.item];
+	[self.context save:&err];
+	
+	sleep(1);
+	
+	tableOperations = [self.context executeFetchRequest:tableOperationsRequest error:&err];
+	
+	XCTAssertEqual(tableOperations.count, 1, @"Should have one insert operation after the save of TodoItem %@", self.item);
```
phvannor: delete.

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,160 @@
...
+	NSManagedObject *tableOperation = tableOperations.firstObject;
+	NSString *operationTable = [tableOperation valueForKey:@"table"];
+	XCTAssertEqualObjects(operationTable, self.item.entity.name, @"The operation should be associated for the %@ table", self.item.entity.name);
+	
+	NSString *operationItemId = [tableOperation valueForKey:@"itemId"];
+	XCTAssertEqualObjects(operationItemId, originalItemId, @"The operation should be associated for the inserted item with id %@", originalItemId);
+	
+	NSDictionary *properties = [[MSJSONSerializer JSONSerializer] itemFromData:[tableOperation valueForKey:@"properties"] withOriginalItem:nil ensureDictionary:YES orError:&err];
+	
+	XCTAssertEqual([properties[@"type"] integerValue], MSTableOperationDelete, @"Associated operation should be an insert with newly created object");
```
phvannor: We should check the delete has the object in it.

damienpontifex:
The last commit adds the callback we were also talking about in #635 that would enable action on error. Also using this with async unit tests to perform checks once completion of the table operation insert had completed

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,176 @@
...
+- (void)tearDown
+{
+	self.store = nil;
+	
+    [super tearDown];
+}
+
+- (void)testObservingInsertOperation
+{
+	XCTestExpectation *expectation = [self expectationWithDescription:@"Table operation observed"];
```
phvannor: Wow, how did I miss this concept when I upgraded our code to use XCTest from sentest.  Very nice.

sdk/iOS/test/MSManagedObjectObserverTests.m
```diff
@@ -0,0 +1,176 @@
...
+@interface MSManagedObjectObserverTests : XCTestCase
+@property (nonatomic, strong) MSClient *client;
+@property (nonatomic, strong) MSCoreDataStore *store;
+@property (nonatomic, strong) NSManagedObjectContext *context;
+@property (nonatomic, strong) MSManagedObjectObserver *observer;
+
+@property (nonatomic, strong) TodoItem *item;
+@end
+
+@implementation MSManagedObjectObserverTests
```
phvannor: Ideally we should have a test case that results in an error to make sure it surfaces correctly.  Perhaps manually put an insert operation for the item to be inserted already in the operation table, that should cause it to error out.

phvannor: Ideally there would also be a test of a save on an item without ms_version, to confirm it does not cause a sync table action.

sdk/iOS/src/MSManagedObjectObserver.h
```diff
@@ -14,4 +17,7 @@
 
 - (instancetype) initWithClient:(MSClient *)client;
 
+/// Block to be called on each operation that will is inserted into MS_TableOperations
```
phvannor: Can you add a note here that not handling errors in this block can result in a mismatch between your client and your server.  I think it will be easy for someone to think of this as optional  step and get their apps in a bad state.  

Long term I'd like to get some storage mechanism that ensures the state has been resolved adequately, but for now that would be good enough.

phvannor:
Looks good, just some minor notes & asks for more test cases.  Really I need to go through and fill a bunch of holes we have as is and set a better example :)  

Thanks for all your work on this so far.

sdk/iOS/test/MSManagedObjectObserverTests.m
damienpontifex: I know. Allowed me to remove a whole heap of dispatch semaphore and dispatch group waits in another project I'm working on, so picked it up quite early :)

damienpontifex:
Happy to get the feedback. Yes, I realise I am probably focusing development on my specific use case as you mentioned about filling in the holes.

damienpontifex:
Another item to note around this is updating quickstarts/documentation. If utilising this observer, I have setup two contexts sharing a common NSPersistentStoreCoordinator.

Main context: used to fetch objects for UI. Also used when initialising client to be passed to observer
Secondary context: used for all calls for sync table pullWithQuery or pushWithCompletion

This avoids the observer noticing changes caused by server operations

phvannor:
I was going to ask you if you ended up needing to do anything extra for the additional context. Eventually need to get that in the .h files etc.  We use doxygen to auto generate the reference docs, so I can go through and add those details in when this is ready to be merged over into Dev.

I'll ask some of our other customers who were interested in this pattern to take a look at this branch and give it try.

PR #641, opened 2015-03-15
MobileServices Client for FirefoxOS

iwate:
I implemented client for FirefoxOS.
Output file is MobileServices.FxOS.js and minified one.
A developer can use following feature of Azure MobileServices by using it.

 * Store data
 * Single sign-on

I implemented push feature too. However can't work without implementing sdk for server.
Your consideration of FirefoxOS support  would be greatly appreciated.


azurecla:
Hi __@iwate__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@iwate__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

sdk/Javascript/src/Push/Push.FxOS.js
```diff
@@ -0,0 +1,413 @@
+// ----------------------------------------------------------------------------
```
phvannor: For this file, does FireFox OS need anything but mdn support? Alternatively, if native or Cordova based JS apps run on FireFox OS , it would be nicer to just add mdn into the Web.js version, and have both the FireFox and Web versions share this file.

sdk/Javascript/src/Platforms/Platform.FxOS.js
```diff
@@ -0,0 +1,277 @@
...
+    // localStorage is not available
+}
+
+var bestAvailableTransport = null;
+var knownTransports = [ // In order of preference
+    require('DirectAjaxTransport'),
+    require('IframeTransport')
+];
+var knownLoginUis = [ // In order of preference
+    require('WebAuthBroker'),
```
phvannor: MozBrowser should be the default correct?  And you don't even need to list the other two?

sdk/Javascript/src/Platforms/Platform.FxOS.js
```diff
@@ -0,0 +1,277 @@
+// ----------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// ----------------------------------------------------------------------------
+
+/// <reference path="..\Generated\MobileServices.DevIntellisense.js" />
+/*global $__fileVersion__:false, $__version__:false */
+
+var _ = require('Extensions');
+var Validate = require('Validate');
+var Promises = require('Promises');
```
phvannor: Does FireFox OS have a built in Promise object like WInJS we could look at using instead of this simple A+ implementation?

sdk/Javascript/src/Platforms/Platform.FxOS.js
```diff
@@ -0,0 +1,277 @@
...
+
+exports.getUserAgent = function () {
+    // Browsers don't allow you to set a custom user-agent in ajax requests. Trying to do so
+    // will cause an exception. So we don't.
+    return null;
+};
+
+exports.getOperatingSystemInfo = function () {
+    return {
+        name: "--",
```
phvannor: Ideally we can populate these on FireFox OS?  See the WinJS version for how we do it for Windows OS apps.

sdk/Javascript/src/Platforms/Platform.FxOS.js
```diff
@@ -0,0 +1,277 @@
...
+        version: "--",
+        architecture: "--"
+    };
+};
+
+exports.getSdkInfo = function () {
+    var isCordovaEnvironment = window && window.cordova && window.cordova.version;
+
+    return {
+        language: isCordovaEnvironment ? "Cordova" : "Web",
```
phvannor: This should never be cordova right?  So we can keep this "Web" only

sdk/Javascript/src/Platforms/Platform.FxOS.js
```diff
@@ -0,0 +1,277 @@
...
+    /// Try to parse an ISO date string.
+    /// </summary>
+    /// <param name="text" type="String">The text to parse.</param>
+    /// <returns type="Date">The parsed Date or null.</returns>
+
+    Validate.isString(text);
+
+    // Check against a lenient regex
+    var matchedDate = isoDateRegex.exec(text);
+    if (matchedDate) {
```
phvannor: It seems like since this all refers to IE, etc, some of this could be simplified for Firefox OS?

phvannor:
How are you testing the FireFox OS JS scripts?  Did you look to see if it was possible to use the current setup to test on FireFox OS?

sdk/Javascript/src/Push/Push.FxOS.js
iwate: In first, I tried to modify Web.js. However parent window object don't access to child window object and  on FirefoxOS app.
Using iframe too.
In result, it need to use MozBrowser (It is like webview on Android app) to work single sign-on. 
Therefore I create new file.

sdk/Javascript/src/Platforms/Platform.FxOS.js
iwate: Yes, it is. MozBrowser should be right default.

sdk/Javascript/src/Platforms/Platform.FxOS.js
iwate: I implemented Platform.FirefoxOS.js on the basis of Platform.Web.js. And, as it was pointed out, it contains a few unnecessary code.
Should I change to simple code by using build in Promise object and send pull request it again?

sdk/Javascript/src/Platforms/Platform.FxOS.js
iwate: Cordova support FirefoxOS now. I think that it should be kept.
https://developer.mozilla.org/ja/Apps/Tools_and_frameworks/Cordova_support_for_Firefox_OS

iwate:
I ran my [sample project](https://github.com/iwate/FxOSTodoSample) on simulator and Fx0. It seemed to work.
If you have a guideline for check, can you tell me what I should do?

sdk/Javascript/src/Push/Push.FxOS.js
phvannor: Sorry I was unclear.  This comment was meant in reference to Push.FxOs.js vs Push.Web.js.  Could we not just add the mdn support to Push.Web.js and use the same file for the FxOS & Web builds?  I don't see any other major difference here. 

Otherwise, does FireFox OS use any of these push systems?  It seems the FireFox OS version should only support mdn?  Why keep apns/gcm in it?

sdk/Javascript/src/Platforms/Platform.FxOS.js
phvannor: Ok

sdk/Javascript/src/Platforms/Platform.FxOS.js
phvannor: If there is a built in promise, and its A+ compliant, then yes we should see if it can be used.

If not, or its alot of work, we can continue to use this version.

phvannor:
Our JS SDK has a unit test suite (https://github.com/Azure/azure-mobile-services/blob/master/README.md#running-the-tests-3), however unlike the build its still requires VS.  

I'm not sure what it would take to make a new project in it for FireFox.  I'm hoping to eventually have this test suite integrate better with grunt & allow running outside of windows.

iwate:
>  I'm hoping to eventually have this test suite integrate better with grunt & allow running outside of windows.

I think so too.

sdk/Javascript/src/Push/Push.FxOS.js
iwate: I think you're right! It is best to add the mdn support to Push.Web.js

sdk/Javascript/src/Platforms/Platform.FxOS.js
iwate: Ok

sdk/Javascript/src/Platforms/Platform.FxOS.js
iwate: Yes, it can. Firefox support ISO-8601 format.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#ECMAScript_5_ISO-8601_format_support

paulbatum:
Mozilla are no longer producing firefox OS phones so I dont expect to ever merge these changes in. Closing.

PR #644, opened 2015-03-17
Fix Issue #413

marianosz:
Delete operations don’t collapse on previous failed insert operations to avoid delete tries to collapse it

Added support to maintain the operation queue on errors.

Added an new “removeTableOperation” method to remove operations with errors from the queue

Added Unit tests to test the new functionality


sdk/android/test/sdk.testapp/src/main/java/com/microsoft/windowsazure/mobileservices/sdk/testapp/test/MobileServiceSyncTableTests.java
```diff
@@ -719,11 +723,394 @@ public void testDeleteCancelsAllWhenInsertIsInQueue() throws Throwable {
         assertEquals(client.getSyncContext().getPendingOperations(), 1);
```
soninaren: name of this function is suggests it returns operations not operationCount,
Name should be getPendingOperationsCount or something similar

PR #662, opened 2015-04-01
iOS: Update synccontext errors and completions

phvannor:
Fix completion blocks not occurring on the specified callback queue.

Surface operation write errors to callers when they occur post item update.

### DIFF
--- a/sdk/iOS/src/MSSyncContext.m
+++ b/sdk/iOS/src/MSSyncContext.m
@@ -36,8 +36,8 @@ static NSOperationQueue *pushQueue_;
     client_ = client;
     operationQueue_ = [[MSOperationQueue alloc] initWithClient:client_ dataSource:self.dataSource];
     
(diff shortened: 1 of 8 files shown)


sdk/iOS/src/MSSyncContext.m
```diff
@@ -126,7 +125,9 @@ -(void) syncTable:(NSString *)table
     
     if (error) {
         if (completion) {
-            completion(nil, error);
+            [self.callbackQueue addOperationWithBlock:^{
+                completion(nil, error);
+            }];
```
brettsam: Can our unit tests verify these are all coming back on the correct queue?

brettsam:
Other than my unit test comment, everything else looks good.

phvannor:
Added check of queue into insert tests, and updated push callback mechanism to match.  Since I'm updating the rest (see my other PR), I'll do the same for update/delete/etc in future ones.

brettsam:
Looks good.

PR #690, opened 2015-04-30
Update header import statements

damienpontifex:
Removing explicit #import in favour of @class declarations. Avoids import chaining when testing against a potential OS X implementation

Have built and run unit tests for iOS and all still succeeds. Further to this, I have created a target for an OS X project on my branch here: https://github.com/damienpontifex/azure-mobile-services/tree/OSX-target that requires these changes.

I have tested the further work from this PR to get an OS X target in a very basic application, just comparing authentication using ADAL for OS X and then passing the token into the mobile service. 

Would be very interested to get feedback and discussion, especially in an attempt to get this OS X support into the framework.


azurecla:

TTYL, AZPRBOT;

damienpontifex:
@phvannor I have done this off the dev branch, but thinking it may be more favourable to do it off the MobileApp branch if the Mac framework were to have the minimum target as v3.0

Thoughts/feedback?

phvannor:
Changes to MobileApp only are more flexible since its in Beta, also slightly easier since I don't have to marshall them from Dev -> Master -> MobileApp. For adding a new platform/etc, it will be easier for me to take if the work goes only to the MobileApp branch.

Also, make sure you put the new file into the build phase sections so the standalone FW build works as well.  I had to make that fix in my PR for the context branch as I forgot to do that when I added the new files.

PR #692, opened 2015-05-04
Mobile app osx

damienpontifex:
As PR #690 but against MobileApp branch for better dev process.

Removing explicit #import in favour of @class declarations. Avoids import chaining when testing against a potential OS X implementation

Have built and run unit tests for iOS and all still succeeds. Further to this, I have created a target for an OS X project on my branch here: https://github.com/damienpontifex/azure-mobile-services/tree/MobileApp-OSX-target that requires these changes.

I have tested the further work from this PR to get an OS X target in a very basic application, just comparing authentication using ADAL for OS X and then passing the token into the mobile service.

Would be very interested to get feedback and discussion, especially in an attempt to get this OS X support into the framework.


azurecla:

TTYL, AZPRBOT;

damienpontifex:
@phvannor updated changes against MobileApp branch as discussed.

This PR is just the changes against the iOS target, but on the other OS X target branch mentioned, I have compiled all of these against a Mac framework target with success. Will create a PR that includes that further commit as well if these all seem good originally.

sdk/iOS/src/WindowsAzureMobileServices.h
```diff
@@ -19,6 +19,32 @@
 #import "MSCoreDataStore.h"
 #import "MSPush.h"
 #import "MSDateOffset.h"
+#import "MSQueryResult.h"
```
phvannor: Why did you add all the files here?  It should be just public facing ones.

sdk/iOS/src/MSUserAgentBuilder.m
```diff
@@ -34,15 +36,19 @@ +(NSString *) userAgent
...
     OS = unknownValue;
     OSversion = unknownValue;
-#else
+#elif TARGET_OS_IPHONE
     UIDevice *currentDevice = [UIDevice currentDevice];
     model = [currentDevice model];
     OS = [currentDevice systemName];
     OSversion = [currentDevice systemVersion];
+#else
+	OSversion = [[NSProcessInfo processInfo] operatingSystemVersionString];
```
phvannor: nit: whitespace

sdk/iOS/src/BlockDefinitions.h
```diff
@@ -0,0 +1,76 @@
...
+
+
+// Callback for connections. If there was an error, the |error| will be non-nil.
+// If there was not an error, the |response| will be non-nil, but
+// the |data| may or may not be nil depending on if the response had content.
+typedef void (^MSResponseBlock)(NSHTTPURLResponse *response,
+								NSData *data,
+								NSError *error);
+
+#pragma mark * MSPushHttp
```
phvannor: delete these, the MobileApp Push doesn't use them now.

sdk/iOS/src/WindowsAzureMobileServices.h
damienpontifex: When building against the Mac framework target, it gave me a warning about no includes for public header files. These were the full list of public headers. Reviewing now, this list is probably smaller. I can review the public headers from the iOS static library and update.

phvannor:
Did you run a simple test on the fw file output via the build script to ensure its still happy?  If so, I think this looks ok.

sdk/iOS/src/MSUserAgentBuilder.m
damienpontifex: In Xcode this extra whitespace doesn't show up. It's just a single tab

damienpontifex:
Yes build against the static library and Framework aggregate target both succeed.

phvannor:
ok this looks good. I tried cloning your MobileApp OSX Target branch and hit a number of issues?  Is that still in progress?

damienpontifex:
Yes it's still a work in progress. Xcode seems to be a lot more picky about exposure of headers and other items: especially when building the framework as a module as well. Continuing to work on that branch/target

damienpontifex:
@phvannor the OS X target branch is building fine on my local machine now. Could you try and let me know if you have any issues?

Unit tests against the OS X target aren't currently implemented because of the module compilation and public exposure of headers with such a build causing some issues. Still working through how that will work.

PR #697, opened 2015-05-12
Make Id column operations case insensitive

shrishrirang:
- Added few UTs
- Also, fixed a table URL UT broken by an earlier commit.

### DIFF
--- a/sdk/iOS/src/MSOperationQueue.m
+++ b/sdk/iOS/src/MSOperationQueue.m
@@ -60,7 +60,7 @@
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:syncTable];
     
     if (item) {
-        query.predicate = [NSPredicate predicateWithFormat:@"(table == %@) AND (itemId == %@)", table, item];
+        query.predicate = [NSPredicate predicateWithFormat:@"(table == %@) AND (itemId ==[c] %@)", table, item];
     } else {
         query.predicate = [NSPredicate predicateWithFormat:@"(table == %@)", table];
     }
--- a/sdk/iOS/test/MSCoreDataStoreTests.m
+++ b/sdk/iOS/test/MSCoreDataStoreTests.m
@@ -58,12 +58,12 @@ NSString * const TableName = @"TodoItem";
     // Seed data to be inserted in the table
     NSString * const originalItemId = @"itemid";
     NSString * const originalText = @"original text";
-    NSArray * const originalItems = @[@{@"id":originalItemId, @"text":originalText}];
+    NSArray * const originalItems = @[ @{ @"id":originalItemId, @"text":originalText } ];
 
     // Data to be used for updating the table. The Id differs from the Id of the seed data only in case
     NSString * const updatedItemId = @"ITEMID";
     NSString * const updatedText = @"updated text";
-    NSArray * const updatedItems = @[@{@"id":updatedItemId, @"text":updatedText}];
+    NSArray * const updatedItems = @[ @{ @"id":updatedItemId, @"text":updatedText } ];
 
     // Populate the table
     [self.store upsertItems:originalItems table:TableName orError:&error];
@@ -80,7 +80,7 @@ NSString * const TableName = @"TodoItem";
 
     // Verify the original row got updated
     item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
-    XCTAssertTrue([item[@"id"] isEqualToString:updatedItemId], @"Incorrect item id"); //ttodoshrirs correct
+    XCTAssertTrue([item[@"id"] isEqualToString:updatedItemId], @"Incorrect item id");
     XCTAssertTrue([item[@"text"] isEqualToString:updatedText], @"Incorrect text");
 
     // Delete the row using an ID with different case
--- a/sdk/iOS/test/MSCoreDataStoreTests.m
+++ b/sdk/iOS/test/MSCoreDataStoreTests.m
@@ -64,6 +64,9 @@ NSString * const TableName = @"TodoItem";
     NSString * const updatedItemId = @"ITEMID";
     NSString * const updatedText = @"updated text";
     NSArray * const updatedItems = @[ @{ @"id":updatedItemId, @"text":updatedText } ];
+    
+    // Initialize table
+    MSSyncTable *syncTable = [[MSSyncTable alloc] initWithName:TableName client:self.client];
 
     // Populate the table
     [self.store upsertItems:originalItems table:TableName orError:&error];
@@ -79,22 +82,21 @@ NSString * const TableName = @"TodoItem";
     XCTAssertNil(error, @"upsert failed: %@", error.description);
 
     // Verify the original row got updated
-    item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
-    XCTAssertTrue([item[@"id"] isEqualToString:updatedItemId], @"Incorrect item id");
-    XCTAssertTrue([item[@"text"] isEqualToString:updatedText], @"Incorrect text");
+    MSQuery *query = [[MSQuery alloc] initWithSyncTable:syncTable predicate:nil];
+    MSSyncContextReadResult *result = [self.store readWithQuery:query orError:&error];
+    XCTAssertNil(error, @"readWithQuery: failed: %@", error.description);
+    XCTAssertEqual(result.items.count, 1, "Expected exactly one item in the table");
+    XCTAssertTrue([result.items[0][@"id"] isEqualToString:updatedItemId], @"Incorrect item id. Did the query return wrong item?");
+    XCTAssertTrue([result.items[0][@"text"] isEqualToString:updatedText], @"Incorrect text. Did the query return wrong item?");
 
     // Delete the row using an ID with different case
     [self.store deleteItemsWithIds:@[originalItemId] table:TableName orError:&error];
     XCTAssertNil(error, @"deleteItemsWithIds: failed: %@", error.description);
 
     // Read to ensure the item is actually gone
-    item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
-    XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
-    XCTAssertNil(item, @"item should have been nil");
-
-    item = [self.store readTable:TableName withItemId:updatedItemId orError:&error];
-    XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
-    XCTAssertNil(item, @"item should have been nil");
+    result = [self.store readWithQuery:query orError:&error];
+    XCTAssertNil(error, @"readWithQuery: failed: %@", error.description);
+    XCTAssertEqual(result.items.count, 0, "Expected the table to be empty");
 }
 
 
--- a/sdk/iOS/test/MSCoreDataStoreTests.m
+++ b/sdk/iOS/test/MSCoreDataStoreTests.m
@@ -105,17 +105,17 @@ NSString * const TableName = @"TodoItem";
     NSError *error;
     NSArray *testArray = @[@{@"id":@"ABC", @"text": @"test1", @"__version":@"APPLE"}];
     
-    [self.store upsertItems:testArray table:@"TodoItem" orError:&error];
+    [self.store upsertItems:testArray table:TableName orError:&error];
     XCTAssertNil(error, @"upsert failed: %@", error.description);
     
-    NSDictionary *item = [self.store readTable:@"TodoItem" withItemId:@"ABC" orError:&error];
+    NSDictionary *item = [self.store readTable:TableName withItemId:@"ABC" orError:&error];
     XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
     XCTAssertNotNil(item, @"item should not have been nil");
     XCTAssertTrue([item[@"id"] isEqualToString:@"ABC"], @"Incorrect item id");
     XCTAssertNotNil(item[MSSystemColumnVersion], @"__version was missing");
     XCTAssertNil(item[@"ms_version"], @"__version was missing");
     
-    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:@"TodoItem" client:self.client];
+    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:TableName client:self.client];
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoItem predicate:nil];
     
     MSSyncContextReadResult *result = [self.store readWithQuery:query orError:&error];
@@ -136,16 +136,16 @@ NSString * const TableName = @"TodoItem";
                             @{@"id":@"B", @"text": @"test2"},
                             @{@"id":@"C", @"text": @"test3"}];
     
-    [self.store upsertItems:testArray table:@"TodoItem" orError:&error];
+    [self.store upsertItems:testArray table:TableName orError:&error];
     XCTAssertNil(error, @"upsert failed: %@", error.description);
     
-    NSDictionary *item = [self.store readTable:@"TodoItem" withItemId:@"B" orError:&error];
+    NSDictionary *item = [self.store readTable:TableName withItemId:@"B" orError:&error];
     XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
     XCTAssertNotNil(item);
     XCTAssertEqualObjects(item[@"id"], @"B");
     XCTAssertEqualObjects(item[@"text"], @"test2");
     
-    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:@"TodoItem" client:self.client];
+    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:TableName client:self.client];
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoItem predicate:nil];
     
     MSSyncContextReadResult *result = [self.store readWithQuery:query orError:&error];
@@ -267,7 +267,7 @@ NSString * const TableName = @"TodoItem";
     
     [self populateTestData];
     
-    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:@"TodoItem" client:self.client];
+    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:TableName client:self.client];
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoItem predicate:nil];
     query.predicate = [NSPredicate predicateWithFormat:@"text == 'test3'"];
     
@@ -284,7 +284,7 @@ NSString * const TableName = @"TodoItem";
     
     [self populateTestData];
     
-    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:@"TodoItem" client:self.client];
+    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:TableName client:self.client];
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoItem predicate:nil];
     query.fetchLimit = 1;
     query.includeTotalCount = YES;
@@ -302,7 +302,7 @@ NSString * const TableName = @"TodoItem";
     
     [self populateTestData];
     
-    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:@"TodoItem" client:self.client];
+    MSSyncTable *todoItem = [[MSSyncTable alloc] initWithName:TableName client:self.client];
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoItem predicate:nil];
     [query orderByAscending:@"sort"];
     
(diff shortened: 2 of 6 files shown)


sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,6 +50,54 @@ -(void)testInit
     XCTAssertNotNil(self.store, @"store creation failed");
 }
 
+// Tests MSCoreDataStore Read, Update and Delete operations are case insensitive for the Id column.
+-(void)testReadUpdateDelete_CaseInsensitiveIdOperations
+{
+    NSError *error;
+
+    // Seed data to be inserted in the table
+    NSString * const originalItemId = @"itemid";
```
phvannor: remove const here, they aren't needed

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,6 +50,54 @@ -(void)testInit
...
+    NSString * const originalText = @"original text";
+    NSArray * const originalItems = @[@{@"id":originalItemId, @"text":originalText}];
```
phvannor: nit formatting use spaces like so: @[ @{ @{ @"id" : val, ... } } ];

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,6 +50,54 @@ -(void)testInit
...
+    XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
+    XCTAssertNotNil(item, @"item should have been found in the table");
+
+    // Update the item using ID with different case
+    [self.store upsertItems:updatedItems table:TableName orError:&error];
+    XCTAssertNil(error, @"upsert failed: %@", error.description);
+
+    // Verify the original row got updated
+    item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
+    XCTAssertTrue([item[@"id"] isEqualToString:updatedItemId], @"Incorrect item id"); //ttodoshrirs correct
```
phvannor: see your comment

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,6 +50,54 @@ -(void)testInit
...
+    // Delete the row using an ID with different case
+    [self.store deleteItemsWithIds:@[originalItemId] table:TableName orError:&error];
+    XCTAssertNil(error, @"deleteItemsWithIds: failed: %@", error.description);
+
+    // Read to ensure the item is actually gone
+    item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
+    XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
+    XCTAssertNil(item, @"item should have been nil");
+
+    item = [self.store readTable:TableName withItemId:updatedItemId orError:&error];
```
phvannor: isn't this portion not needed?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1821,6 +1821,47 @@ -(void) testIncrementalPullWithInsert
...
+    MSClient *filteredClient = [client clientWithFilter:filter];
+    MSSyncTable *todoTable = [filteredClient syncTableWithName:@"TodoItem"];
+    MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoTable];
+
+    // Perform an insert operation in the middle of a pull operation
+    ((MSTestFilter *)filter.testFilters[0]).onInspectRequest = ^(NSURLRequest *request) {
+
+        // Insert an item with Id differing from that of the pulled item only in case
+        NSDictionary *item = @{ @"id": @"ONE", @"name":@"fifth item" };
+        MSTestWaiter *insertWaiter = [MSTestWaiter new];
```
phvannor: test waiter should be replaced with the built in XC test expectation paradigm.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1821,6 +1821,47 @@ -(void) testIncrementalPullWithInsert
...
+    
+        [todoTable insert:item completion:^(NSDictionary *item, NSError *error) {
```
phvannor: this isn't going to simulate the standard pull merge into an id logic. I thought the client will detect the record changed and not even try to upsert now.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1821,6 +1821,47 @@ -(void) testIncrementalPullWithInsert
...
+
+    done = NO;
+    [todoTable pullWithQuery:query queryId:@"test-1" completion:^(NSError *error) {
+        XCTAssertNil(error, @"Error found: %@", error.description);
+        XCTAssertEqual(offline.upsertCalls, 4);
+        XCTAssertEqual(offline.upsertedItems, 4);
+        NSLog(@"done pull");
+        done = YES;
+    }];
+    XCTAssertTrue([self waitForTest:30.0], @"Test timed out.");
```
phvannor: I would expect a check on the raw SQL storage layer, or something to confirm there is only 1 record in the db for this case.

sdk/iOS/test/MSUserAgentBuilderTests.m
```diff
@@ -33,7 +33,7 @@ -(void) testURLForTable
 {
     NSString *userAgent = [MSUserAgentBuilder userAgent];
     
-    XCTAssertTrue([userAgent isEqualToString:@"ZUMO/2.1 (lang=objective-c; os=--; os_version=--; arch=iOSSimulator; version=2.1.0)"],
+    XCTAssertTrue([userAgent isEqualToString:@"ZUMO/3.0 (lang=objective-c; os=--; os_version=--; arch=iOSSimulator; version=3.0.0)"],
```
phvannor: did we say this shouldn't go to dev as well?  Thinking on this changeset, it needs to go into dev as well.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: We don't want someone to change TableName - we want it to be readonly.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: done

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Removed

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Yes, we can argue it is redundant at this stage, but it makes the test more exhaustive. That way we don't tightly couple the test with the existing implementation.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: I was not aware of XCTestExpectation! :+1:  I actually implemented something similar and was planning to send a PR from another branch.
XCTestExpectation does not seem to support nested fulfills/waits. Here we want to wait for the pull operation to complete and independently wait for the insert operation to complete. Do you know if there is a way to achieve this?

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: If the insert is performed before the pull operation checks for inserts, there won't be a merge - instead there will be a push + pull. Here we are inserting after the check is done thus doing the operation queue item merge.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSCoreDataStoreTests.m
phvannor: The same holds true for almost all the variables in these tests and this is not the style we've been following to date, so I'd prefer to match current style all else being equal.  If you wanted to define static consts at the top of the file so all these tests could share 2 different cased Ids that would be a cleaner approach.

sdk/iOS/test/MSUserAgentBuilderTests.m
shrishrirang: yeah right. Let's continue to use this PR for review. Once the changes look ok, I will send a fresh PR for dev.

sdk/iOS/test/MSSyncTableTests.m
phvannor: Not sure I follow, why wouldn't 2 expectations that you are waiting on work?

sdk/iOS/test/MSSyncTableTests.m
phvannor: There should be 2 tests here I believe:
1 - Pull -> Get id of "ABC" -> Verify its upsert into "abc" works fine.  This is the test I expected from the name of the method.  If you want to say this case is covered by the tests directly in MSCoreDataStore and not needed, I can accept that.

2 - A test of the merge when a pending operation is queued during the pull, which is what we this appears to be doing.  I don't think the inc pull matters in this flow.  This would just be testPullKeepsPendingChangesWithDifferentCasedId or some such.  This tests that the record retrieved from the server is discarded when there is already a pending op in the queue and not written back to disk and the pending op is merged correctly back in.

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,6 +50,54 @@ -(void)testInit
...
+    NSDictionary *item = [self.store readTable:TableName withItemId:updatedItemId orError:&error];
+    XCTAssertNil(error, @"readTable:withItemId: failed: %@", error.description);
+    XCTAssertNotNil(item, @"item should have been found in the table");
+
+    // Update the item using ID with different case
+    [self.store upsertItems:updatedItems table:TableName orError:&error];
+    XCTAssertNil(error, @"upsert failed: %@", error.description);
+
+    // Verify the original row got updated
+    item = [self.store readTable:TableName withItemId:originalItemId orError:&error];
```
phvannor: since readWithItemId returns the firstObject, I don't think this guarantees you updated the original record (it relies upon the internal implementation of local store at least).  We probably need to run readWithQuery, and verify only 1 result for that instead.  (And that result has the new data)

shrishrirang: Nice catch! :+1:

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Yes, ideally all strings that are being used repeatedly should be defined as constants. I don't think it is about style. I would like to retain the const variable as it is a good coding practice.

I have changed all uses of @"TodoItem" in this file to use the const variable now. Use of cased ID constant strings is restricted to only one function as of now, so those consts are local to that function (and not defined at the class level)

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -20,6 +20,9 @@ @interface MSCoreDataStoreTests : XCTestCase {
 
 @implementation MSCoreDataStoreTests
 
+// Table used for running tests
+NSString * const TableName = @"TodoItem";
```
phvannor: static

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,22 +50,72 @@ -(void)testInit
...
 }
 
+// Tests MSCoreDataStore Read, Update and Delete operations are case insensitive for the Id column.
+-(void)testReadUpdateDelete_CaseInsensitiveIdOperations
+{
+    NSError *error;
+
+    // Seed data to be inserted in the table
+    NSString * const originalItemId = @"itemid";
+    NSString * const originalText = @"original text";
```
phvannor: Just put the text in the array definition.

sdk/iOS/test/MSCoreDataStoreTests.m
phvannor: Style guide is *const, no space.  If we are keeping them, when are we proposing to use them?  I notice the syncTable and the query are not using it.  Its a lil ambiguous and it doesn't line up with patterns we've used throughout the test suite.

sdk/iOS/test/MSCoreDataStoreTests.m
```diff
@@ -47,22 +50,72 @@ -(void)testInit
...
+    NSArray * const originalItems = @[ @{ @"id":originalItemId, @"text":originalText } ];
+
+    // Data to be used for updating the table. The Id differs from the Id of the seed data only in case
+    NSString * const updatedItemId = @"ITEMID";
+    NSString * const updatedText = @"updated text";
+    NSArray * const updatedItems = @[ @{ @"id":updatedItemId, @"text":updatedText } ];
```
phvannor: nit style: @"id" : updatedItemId (always use a space to either side of the : in the definition)

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -2310,6 +2360,27 @@ -(void) testPurgeWithPendingOperationsFails
...
 
+// Tests operation queue querying is case insensitive for Id column
+-(void)testOperationQueueQueryingForIdCaseInsensitivity
+{
+    MSSyncTable *todoTable = [[MSSyncTable alloc] initWithName:@"TodoItem" client:client];
+    
+    NSDictionary *item = @{ @"id":@"itemid", @"name":@"some name" };
+    
+    // Perform an insert operation
+    done = NO;
```
phvannor: nit: use XCTestExpectation.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -2310,6 +2360,27 @@ -(void) testPurgeWithPendingOperationsFails
     XCTAssertTrue([self waitForTest:0.1], @"Test timed out.");
 }
 
+// Tests operation queue querying is case insensitive for Id column
+-(void)testOperationQueueQueryingForIdCaseInsensitivity
+{
+    MSSyncTable *todoTable = [[MSSyncTable alloc] initWithName:@"TodoItem" client:client];
+    
+    NSDictionary *item = @{ @"id":@"itemid", @"name":@"some name" };
```
phvannor: nit: style

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -2310,6 +2360,27 @@ -(void) testPurgeWithPendingOperationsFails
...
+    [todoTable insert:item completion:^(NSDictionary *item, NSError *error) {
+        XCTAssertNil(error, @"error should have been nil.");
+        done = YES;
+    }];
+    XCTAssertTrue([self waitForTest:0.1], @"Test timed out.");
+    
+    // Query the operation queue for the inserted item using a different case for the Id
+    NSArray *items = [client.syncContext.operationQueue getOperationsForTable:todoTable.name item:@"ITEMID"];
+    
+    XCTAssertEqual([items count], 1, "Expected to find 1 item in the operation queue");
```
phvannor: nit: items.count

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1821,6 +1821,56 @@ -(void) testIncrementalPullWithInsert
...
+    
+        return request;
+    };
+
+    XCTestExpectation *pullExpectation = [self expectationWithDescription:@"table pull"];
+    [todoTable pullWithQuery:query queryId:@"test-1" completion:^(NSError *error) {
+        XCTAssertNil(error, @"Error found: %@", error.description);
+        XCTAssertEqual(offline.upsertCalls, 4);
+        XCTAssertEqual(offline.upsertedItems, 4);
+        NSLog(@"done pull");
```
phvannor: nit: remove logging.

sdk/iOS/src/MSOperationQueue.m
```diff
@@ -60,7 +60,7 @@ -(NSArray *) getOperationsForTable:(NSString *)table item:(NSString *)item
     MSQuery *query = [[MSQuery alloc] initWithSyncTable:syncTable];
     
     if (item) {
-        query.predicate = [NSPredicate predicateWithFormat:@"(table == %@) AND (itemId == %@)", table, item];
+        query.predicate = [NSPredicate predicateWithFormat:@"(table == %@) AND (itemId ==[c] %@)", table, item];
```
phvannor: table should probably also be case insensitive comparison as well.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: The same strings are used in multiple places. Repeating string literals is a potential maintenance pain. Are there advantages of not defining the variables once and reusing multiple times?

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Ok, done. Looks like there are many more instances that are not following this convention. I am not changing them.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/src/MSOperationQueue.m
shrishrirang: There are cons of treating table name in a case insensitive manner. If we think that is needed let us take it as a separate work item.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Didn't realize defining a variable inside an implementation block does not make it an class member. Added static now.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Updated formatting to *const now.
We should try to be const correct for variables that have scope wider than a function. And looks like we have been doing it for global string literals in test as well as product code, which is good (try searching for static nsstring in test code).
Although using const for syncTable and query would be extremely ideal, it is ok to be lenient as it has a restricted scope (function) and less chances of being modified unintentionally. 
Also, IMHO using const more liberally can only be good.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: I need waitForExpectationA and waitForExpectationB calls to be made concurrently. [expectationA fulfill] should unblock waitForExpectationA and [expectationB fulfill] to unblock waitForExpectationB. 
However, waitForExpectation operates on XCTestCase and not on an instance of XCTestExpectation. Besides only 1 call to waitForExpectation is allowed at a time as per Apple's documentation.

sdk/iOS/test/MSCoreDataStoreTests.m
phvannor: originalText is only used in 1 place, defining originalItems

sdk/iOS/src/MSOperationQueue.m
phvannor: Yes, it will need to be done.  I'm not sure what cons you see, but lets get it on the planning radar then.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Right, MSCoreDataStore tests takes care of case 1. This test focuses on case 2. Renamed the method to make it more obvious.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: As discussed, using const only for globals

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: These are not consts anymroe

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: As discussed, using XCTestExpectation as much as possible and continuing to use one instance of MSTestFilter. Will figure out some way to address this kind of asynchronous scenario.

sdk/iOS/src/MSOperationQueue.m
shrishrirang: Ok, added to backlog.

sdk/iOS/test/MSCoreDataStoreTests.m
shrishrirang: Ok removed it.

PR #703, opened 2015-05-17
Returning push and pull operations

damienpontifex:
The user can use the returned operations to cancel, or add as dependencies etc as per normal NSOperation usage.

### DIFF
--- a/sdk/iOS/src/MSSyncContext.h
+++ b/sdk/iOS/src/MSSyncContext.h
@@ -9,6 +9,7 @@
 @class MSSyncContext;
 @class MSTableOperation;
 @class MSSyncContextReadResult;
+@class MSQueuePushOperation;
 
 /// The MSSyncContextDelegate allows for customizing the handling of errors, conflicts, and other
 /// conditions that may occur when syncing data between the device and the mobile service.
@@ -101,7 +102,7 @@
 @property (nonatomic, readonly) NSUInteger pendingOperationsCount;
 
 /// Executes all current pending operations on the queue
-- (void) pushWithCompletion:(MSSyncBlock)completion;
+- (MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion;
 
 /// Specifies the delegate that will be used in the resolution of syncing issues
 @property (nonatomic, strong) id<MSSyncContextDelegate> delegate;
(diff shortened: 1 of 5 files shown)


azurecla:

TTYL, AZPRBOT;

phvannor:
I actually would prefer this on Dev and then I'll bring over to MobileApp.

sdk/iOS/src/MSSyncContext.h
```diff
@@ -101,7 +102,7 @@
 @property (nonatomic, readonly) NSUInteger pendingOperationsCount;
 
 /// Executes all current pending operations on the queue
-- (void) pushWithCompletion:(MSSyncBlock)completion;
+- (MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion;
```
phvannor: I'm not sure how I feel about making MSQueuePush/PullOperation classes entirely public as is.  (And exposing that they are NSOperations)  Are there benefits to exposing the NSOperation itself beyond the cancel call?  I'll need to look at the public .h file to review what API is now being exposed.  Also, the .h files need added in the main list for the FW & the public list for the build?

damienpontifex: I don't generally need to cancel, but you could.

My main use case would be adding other operations as dependencies when calling multiple `pullWithQuery`. i.e. I loop through tables and pull on each. If the actual operation was returned, I then have another NSOperation instance which I can add all of these pulls as dependencies of said operation.

Currently they are project scoped. I get around this, as I am importing the source via cocoa pods and so can access all the headers. If you agree that returning the operations will be acceptable, I can make those headers public.

damienpontifex:
@phvannor if you are happy to make the operations public and return them as suggested, I can close this and create the change on the dev branch. Can you confirm your thoughts on their being returned here?

phvannor:
@damienpontifex you make a good case for exposing the NSOperations.  I'm OK with exposing them to the caller.  (probably should expose the purge one as well to be complete)  Looks like the .h files are pretty empty, so should be fine to move to the public sections.  Ideally add a unit test case to make sure it returns a valid NSOperation you can do your workflow with.

phvannor:
#709 replaces this, we will merge Dev into MobileApp post accepting 709.

Issue #669, opened 2015-04-08
[iOS] Add Objective-C language enhancements for nullability


damienpontifex:
In the [Objective-C Language Enhancements](https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW10) for Xcode 6.3 release notes: parameters, return types and properties can now be annotated with their nullability. 

I have done an example on MSClient.h with this commit damienpontifex/azure-mobile-services@17fc313

Can this be implemented among the classes for the iOS SDK?

phvannor:
Looks feasible, from what I can tell it shouldn't impact users of the fw on older platforms.

damienpontifex:
I believe it is all just compiler checking (mainly for Swift, but also with ObjC benefits) so shouldn't effect older target platforms. Would require developers to be on Xcode 6.3 or above.

damienpontifex:
Closing with integration of #715

PR #704, opened 2015-05-17
initial work on annotating nullable types for better Swift compatibility

damienpontifex:
Adding nullability annotations to MSClient relating to #669 . Only added to this class as a first stage and to receive feedback on these initial annotations and their usage.

I have run trials with this in my own Swift application and they seem be valid null/nonnull parameters. If these seem good and are accepted, can continue to markup other public classes in this trend.

### DIFF
--- a/sdk/iOS/src/MSClient.h
+++ b/sdk/iOS/src/MSClient.h
@@ -143,19 +143,19 @@
 -(void)loginWithProvider:(nonnull NSString *)provider
               controller:(nonnull UIViewController *)controller
                 animated:(BOOL)animated
-              completion:(nonnull MSClientLoginBlock)completion;
+              completion:(nullable MSClientLoginBlock)completion;
 
 /// Returns an MSLoginController that can be used to log in the current
 /// end user with the given provider.
 -(nonnull MSLoginController *)loginViewControllerWithProvider:(nonnull NSString *)provider
-                                 completion:(nonnull MSClientLoginBlock)completion;
+                                 completion:(nullable MSClientLoginBlock)completion;
 #endif
 
 /// Logs in the current end user with the given provider and the given token for
 /// the provider.
 -(void)loginWithProvider:(nonnull NSString *)provider
                    token:(nonnull NSDictionary *)token
-              completion:(nonnull MSClientLoginBlock)completion;
+              completion:(nullable MSClientLoginBlock)completion;
 
 /// Logs out the current end user.
 -(void)logout;
@@ -169,10 +169,6 @@
 
 /// Returns an MSTable instance for a table with the given name.
 -(nonnull MSTable *)tableWithName:(nonnull NSString *)tableName;
-
-/// Old method to return an MSTable instance for a table with the given name.
-/// This has been deprecated. Use tableWithName:
--(nonnull MSTable *)getTable:(nonnull NSString *)tableName __deprecated;
 
 /// Returns an MSSyncTable instance for a table with the given name.
 -(nonnull MSSyncTable *)syncTableWithName:(nonnull NSString *)tableName;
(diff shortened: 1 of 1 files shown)


azurecla:

TTYL, AZPRBOT;

sdk/iOS/src/MSClient.h
```diff
@@ -140,22 +140,22 @@
 #if TARGET_OS_IPHONE
 /// Logs in the current end user with the given provider by presenting the
 /// MSLoginController with the given controller.
--(void)loginWithProvider:(NSString *)provider
-              controller:(UIViewController *)controller
+-(void)loginWithProvider:(nonnull NSString *)provider
+              controller:(nonnull UIViewController *)controller
                 animated:(BOOL)animated
-              completion:(MSClientLoginBlock)completion;
+              completion:(nonnull MSClientLoginBlock)completion;
```
phvannor: this can be null.

sdk/iOS/src/MSClient.h
```diff
@@ -140,22 +140,22 @@
...
 
 /// Returns an MSLoginController that can be used to log in the current
 /// end user with the given provider.
--(MSLoginController *)loginViewControllerWithProvider:(NSString *)provider
-                                 completion:(MSClientLoginBlock)completion;
+-(nonnull MSLoginController *)loginViewControllerWithProvider:(nonnull NSString *)provider
+                                 completion:(nonnull MSClientLoginBlock)completion;
```
phvannor: same for all the completion blocks.

sdk/iOS/src/MSClient.h
```diff
@@ -168,14 +168,14 @@
 /// @{
 
 /// Returns an MSTable instance for a table with the given name.
--(MSTable *)tableWithName:(NSString *)tableName;
+-(nonnull MSTable *)tableWithName:(nonnull NSString *)tableName;
 
 /// Old method to return an MSTable instance for a table with the given name.
 /// This has been deprecated. Use tableWithName:
--(MSTable *)getTable:(NSString *)tableName __deprecated;
+-(nonnull MSTable *)getTable:(nonnull NSString *)tableName __deprecated;
```
phvannor: might as well just delete this guy finally.

sdk/iOS/src/MSClient.h
```diff
@@ -186,21 +186,21 @@
...
 /// response content will be treated as JSON.
--(void)invokeAPI:(NSString *)APIName
-            body:(id)body
-      HTTPMethod:(NSString *)method
-      parameters:(NSDictionary *)parameters
-         headers:(NSDictionary *)headers
-      completion:(MSAPIBlock)completion;
+-(void)invokeAPI:(nonnull NSString *)APIName
+            body:(nullable id)body
+      HTTPMethod:(nonnull NSString *)method
```
phvannor: i believe this is allowed to be null, and if so, defaults to POST

sdk/iOS/src/MSClient.h
```diff
@@ -140,22 +140,22 @@
 #if TARGET_OS_IPHONE
 /// Logs in the current end user with the given provider by presenting the
 /// MSLoginController with the given controller.
--(void)loginWithProvider:(NSString *)provider
-              controller:(UIViewController *)controller
+-(void)loginWithProvider:(nonnull NSString *)provider
+              controller:(nonnull UIViewController *)controller
```
phvannor: did you check this one?  I'm not sure if this is required or not off the top of my head

damienpontifex: If you don't provide the view controller, it cannot present the login controller. Hence why I chose nonnull on that one.

On the provider: the request generated in MSLogin appends login/<provider> to the gateway url. I would've though it wouldn't make sense to provide nil as the gateway would reject it as not knowing how to handle it? But I don't have the specific source code to confirm that for the backend.

sdk/iOS/src/MSClient.h
damienpontifex: True, it does in MSAPIRequest

PR #706, opened 2015-05-18
Mobile app osx target

damienpontifex:
Compiling for OS X framework target. 

Only missing components are the MSLoginView and MSLoginController as they use UI specific classes. Thus the only login methods are currently with existing token. This could be rectified in the future to construct the equivalent classes using NSView and NSViewController subclasses for the equivalent functionality.

I have also worked on a quick start to test the basic pull, insert, update functionality on another branch at [MobileApp-Mac-Quickstart](https://github.com/damienpontifex/azure-mobile-services/tree/MobileApp-Mac-Quickstart)


azurecla:

TTYL, AZPRBOT;

damienpontifex:
@phvannor on the Mac quick start, I didn't quite understand what you had done with the build.command files in the other quick starts. Could you outline what would be needed to add these to the mac quick start to build the corresponding target in the framework project?

damienpontifex:
@phvannor is the Mac target being considered for integration in the mobile services framework?

phvannor:
I keep meaning to spend some time on this, but never get a chance.  I did fix up the unit tests and run them and they seemed to pass.  If you can fix the conflicts now, and wire up the tests (missing adding the buildphase copy of the datamodel and none of the tests were checked off to be included in the mac suite) that would let us take it as is.  Official support & release of the fw would be a ways away though.

damienpontifex:
I will have a look at bringing it up to date this week.

damienpontifex:
@phvannor this should now be up to date with the MobileApp branch.

Logic unit tests are integrated passing on all targets: iOS static library, iOS dynamic framework & Mac Dynamic Framework.

Functional tests fail on the dynamic frameworks, but they did before the merge so I assume this is intentional?

phvannor:
Yes, I'm not sure the state of the dynamic framework and test beds yet.   I'll have to see if someone can look into it.  Thanks for the update, I'll try to get this in quickly this time.

sdk/iOS/WindowsAzureMobileServices.xcodeproj/project.pbxproj
```diff
@@ -49,16 +49,16 @@
 		5A4FAC8F1B05E5250027011D /* MSUser.h in Headers */ = {isa = PBXBuildFile; fileRef = E8F33B12161667CE002DD7C6 /* MSUser.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		5A4FAC901B05E52C0027011D /* MSUser.m in Sources */ = {isa = PBXBuildFile; fileRef = E8F33B13161667CE002DD7C6 /* MSUser.m */; };
 		5A4FAC911B05E5300027011D /* MSFilter.h in Headers */ = {isa = PBXBuildFile; fileRef = E8E378FF161E14E800C13F00 /* MSFilter.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		5A4FAC921B05E5360027011D /* MSLocalStorage.h in Headers */ = {isa = PBXBuildFile; fileRef = A24701AD196896F500385DA2 /* MSLocalStorage.h */; };
-		5A4FAC931B05E53B0027011D /* MSLocalStorage.m in Sources */ = {isa = PBXBuildFile; fileRef = A24701AE196896F500385DA2 /* MSLocalStorage.m */; };
+		5A4FAC921B05E5360027011D /* (null) in Headers */ = {isa = PBXBuildFile; };
```
phvannor: this doesn't look quite right (null) ?

sdk/iOS/src/WindowsAzureMobileServices.h
```diff
@@ -11,7 +11,7 @@
 #import "MSError.h"
 #import "MSFilter.h"
 #if TARGET_OS_IPHONE
-#import "MSLoginController.h"
+#import <WindowsAzureMobileServices/MSLoginController.h>
```
phvannor: out of curiosity, why does this need the full path?

sdk/iOS/WindowsAzureMobileServices.xcodeproj/project.pbxproj
phvannor: oh the last merge from master seems to have revived links to removed files, just need removed from App branch.  I can do that quickly.

phvannor:
Added #800 which fixes the pbx, and probably does some of the same stuff you are doing here.  Take a look, and I'll merge that in or let me know if you want to do the same cleanup instead on yours.

sdk/iOS/src/WindowsAzureMobileServices.h
damienpontifex: Not sure to be honest. 

Although if you look at the Apple umbrella headers for their frameworks they list it with the angle brackets and full path e.g. inside `#import <Foundation/Foundation.h>` they have `#import <Foundation/NSArray.h>` so it's probably just a convention for header searching.

I can change it to just the file to be consistent with the others

damienpontifex:
Not sure why the (null) values in the project file existed, but removed and seems fine.

I have included all headers in the iOS dynamic framework target at project level and the one BlockDefinitions.h at public level and added in the reference to the prefix header and the build settings. This seems to cover everything in #800 but let me know if I missed anything.

phvannor:
@damienpontifex can you try running the test suite for OS X?  I'm getting some must import X from module errors in TableOperationInternal, not sure if a bad merge or what is not letting them run.

damienpontifex:
I'm not getting any of these errors. I did a clean, removed Derived data and ran again with no issues.

I also did a fresh clone of the repository in a new location to make sure I had all appropriate files checked in a re-ran with same result.

What is 'X' that you must import? Some system framework?

phvannor:
ok, running fine for me now, must have been bad state, thanks for checking.

PR #707, opened 2015-05-18
[iOS] added property to turn off data store handling table operations

damienpontifex:
Used for custom implementation of data insertion and requiring the sync context to only handle server message creation.

Won't cause any breaking changes to existing clients as the property for the MSCoreDataStore will default to YES. Does allow custom implementations per app for insertion into the data store, as experimented with in the multiple-contexts branch.

### DIFF
--- a/sdk/iOS/src/MSCoreDataStore.h
+++ b/sdk/iOS/src/MSCoreDataStore.h
@@ -25,6 +25,12 @@
 
 /// @}
 
+/// Disables the store from recieving information about the items passed into all sync table
+/// calls (insert, delete, update). If set, the application is responsible for already having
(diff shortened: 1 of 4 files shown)


azurecla:

TTYL, AZPRBOT;

phvannor:
This is just grabbed from the other branch?

damienpontifex:
Yes. But just the control properties.

i.e. leaves the actual implementation of the observer separate. I thought pulling this in wouldn't have any effect on clients that didn't care, but allow some to implement custom insert logic. Similar to what we have done in the multiple-contexts branch, but bringing that forward into the MobileApp so I can keep that implementation in my project while not releasing that feature to the public build in here.

PR #709, opened 2015-05-19
[iOS] Returning operations from sync actions

damienpontifex:
Closes #703 by bringing these changes into dev branch first.

Returns NSOperation types from sync operations to allow the use to use them to react to changes. This could be to cancel, or to add these as dependencies to a trailing operation.

### DIFF
--- a/sdk/iOS/src/MSSyncContext.h
+++ b/sdk/iOS/src/MSSyncContext.h
@@ -110,7 +110,7 @@ typedef void (^MSSyncPushCompletionBlock)(void);
 @property (nonatomic, readonly) NSUInteger pendingOperationsCount;
 
 /// Executes all current pending operations on the queue
-- (MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion;
+- (NSOperation *) pushWithCompletion:(MSSyncBlock)completion;
 
 /// Specifies the delegate that will be used in the resolution of syncing issues
 @property (nonatomic, strong) id<MSSyncContextDelegate> delegate;
--- a/sdk/iOS/src/MSSyncContext.h
+++ b/sdk/iOS/src/MSSyncContext.h
@@ -8,6 +8,7 @@
 
 @class MSQuery;
 @class MSSyncContext;
+@class MSQueuePushOperation;
 
 /// Callback for updates and deletes. If there was an error, the *error* will be non-nil.
 typedef void (^MSSyncBlock)(NSError *error);
@@ -109,7 +110,7 @@ typedef void (^MSSyncPushCompletionBlock)(void);
 @property (nonatomic, readonly) NSUInteger pendingOperationsCount;
 
 /// Executes all current pending operations on the queue
-- (void) pushWithCompletion:(MSSyncBlock)completion;
+- (MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion;
 
 /// Specifies the delegate that will be used in the resolution of syncing issues
 @property (nonatomic, strong) id<MSSyncContextDelegate> delegate;
--- a/sdk/iOS/src/MSSyncContext.h
+++ b/sdk/iOS/src/MSSyncContext.h
@@ -8,7 +8,6 @@
 
 @class MSQuery;
 @class MSSyncContext;
-@class MSQueuePushOperation;
 
 /// Callback for updates and deletes. If there was an error, the *error* will be non-nil.
 typedef void (^MSSyncBlock)(NSError *error);
--- a/sdk/iOS/src/MSSyncContext.m
+++ b/sdk/iOS/src/MSSyncContext.m
@@ -82,7 +82,7 @@ static NSOperationQueue *pushQueue_;
 /// Begin sending pending operations to the remote tables. Abort the push attempt whenever any single operation
 /// recieves an error due to network or authorization. Otherwise operations will all run and all errors returned
 /// to the caller at once.
--(MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion
+-(NSOperation *) pushWithCompletion:(MSSyncBlock)completion
 {
     // TODO: Allow users to cancel operations
     MSQueuePushOperation *push = [[MSQueuePushOperation alloc] initWithSyncContext:self
@@ -431,13 +431,14 @@ static NSOperationQueue *pushQueue_;
 				// For now we just abort the pull if the push failed to complete successfully
 				// Long term we can be smarter and check if our table succeeded
 				if (error) {
+					[pull cancel];
+					
 					if (completion) {
 						[self.callbackQueue addOperationWithBlock:^{
 							completion(error);
 						}];
 					}
 				} else {
-					// Check again if we have new pending ops while we synced, and repeat as needed
 					[pushQueue_ addOperation:pull];
 				}
 			}];
--- a/sdk/iOS/src/MSSyncContext.m
+++ b/sdk/iOS/src/MSSyncContext.m
@@ -11,7 +11,7 @@
 #import "MSQuery.h"
 #import "MSQueryInternal.h"
 #import "MSQueuePushOperation.h"
-#import "MSQueuePullOperation.h"
+#import "MSQueuePullOperationInternal.h"
 #import "MSQueuePurgeOperation.h"
 #import "MSNaiveISODateFormatter.h"
 #import "MSDateOffset.h"
@@ -432,6 +432,7 @@ static NSOperationQueue *pushQueue_;
                 // Long term we can be smarter and check if our table succeeded
                 if (error) {
                     [pull cancel];
+					[pull completeOperation];
                     
                     if (completion) {
                         [self.callbackQueue addOperationWithBlock:^{
--- a/sdk/iOS/src/MSSyncContext.m
+++ b/sdk/iOS/src/MSSyncContext.m
@@ -82,7 +82,7 @@ static NSOperationQueue *pushQueue_;
 /// Begin sending pending operations to the remote tables. Abort the push attempt whenever any single operation
 /// recieves an error due to network or authorization. Otherwise operations will all run and all errors returned
 /// to the caller at once.
--(void) pushWithCompletion:(MSSyncBlock)completion
+-(MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion
 {
     // TODO: Allow users to cancel operations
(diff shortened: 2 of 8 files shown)


azurecla:

TTYL, AZPRBOT;

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,53 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
+	MSQueuePullOperation *pull = [[MSQueuePullOperation alloc] initWithSyncContext:self
+																			 query:query
+																		   queryId:queryId
+																		maxRecords:maxRecords
+																	 dispatchQueue:writeOperationQueue
+																	 callbackQueue:self.callbackQueue
+																		completion:completion];
+	
+	// Before we can pull from the remote, we need to make sure out table doesn't having pending operations
+	NSArray *tableOps = [self.operationQueue getOperationsForTable:query.table.name item:nil];
```
phvannor: Keeping this check on the original queue here will incur a performance penalty in the user's app, as now they are waiting for a db read.  What's the rationale for keeping it here?

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,53 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
+			// Long term we can be smarter and check if our table succeeded
+			if (error) {
+				if (completion) {
+					[self.callbackQueue addOperationWithBlock:^{
+						completion(error);
+					}];
+				}
+			} else {
+				// Check again if we have new pending ops while we synced, and repeat as needed
+				[pushQueue_ addOperation:pull];
```
phvannor: This will be a change in behavior.  We aren't checking the if table is dirty again, just assuming we are good to go on success.  It probably has minimal impact, but the comments on the functions are now wrong.

phvannor:
I expected to see an update to sdk/iOS/src/WindowsAzureMobileServices.h to list the Queue*Operation files, as well as adding the  {ATTRIBUTES = (Public, ) option in the pbxproj file.

damienpontifex:
@shrishrirang that seems like a good solution. I have tested that and seems to work as well.

sdk/iOS/src/MSSyncContext.m
damienpontifex: I couldn't figure out how to recheck if the table was dirty and still utilise the pull operation we returned to the user. 

The only was I cam up with was extracting this dirty/push check into another private method which took the existing pull operation. Thus we could call it again in the callback and pass the contained push operation. Thoughts?

sdk/iOS/src/MSSyncContext.m
damienpontifex: Originally I didn't see a way to return the pull operation from an async queue. Then I moved the pull constructor up to the first expression and I see I can add the async call back in. Will do it.

sdk/iOS/src/MSSyncContext.m
phvannor: I'm OK with the current logic, and only doing 1 push for now, after I wrote this, I ended up adding code to handle pending ops post pull anyways, so its not really necessary to even do the 1 push.  In addtion cancelling the pull NSOperation, should cancel the dependent Push, so this whole section may need reworked later on anyways.  For now, just fix the docs on this function to indicate 1 push will be attempted if pending ops exist for the table.

sdk/iOS/src/MSSyncContext.h
```diff
@@ -8,6 +8,7 @@
 
 @class MSQuery;
 @class MSSyncContext;
+@class MSQueuePushOperation;
```
shrishrirang: Is this change needed now (in this file and other files)?

sdk/iOS/src/MSSyncContext.m
```diff
@@ -82,7 +82,7 @@ -(NSUInteger) pendingOperationsCount
 /// Begin sending pending operations to the remote tables. Abort the push attempt whenever any single operation
 /// recieves an error due to network or authorization. Otherwise operations will all run and all errors returned
 /// to the caller at once.
--(void) pushWithCompletion:(MSSyncBlock)completion
+-(MSQueuePushOperation *) pushWithCompletion:(MSSyncBlock)completion
```
shrishrirang: NSOperation* return type?

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,55 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
-                                                                                maxRecords:maxRecords
-                                                                             dispatchQueue:writeOperationQueue
-                                                                             callbackQueue:self.callbackQueue
-                                                                                completion:completion];
-            
-            
-            [pushQueue_ addOperation:pull];
-        }
-    });
+	MSQueuePullOperation *pull = [[MSQueuePullOperation alloc] initWithSyncContext:self
```
shrishrirang: Minor: Can you fix the formatting?

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,55 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
+				// For now we just abort the pull if the push failed to complete successfully
+				// Long term we can be smarter and check if our table succeeded
+				if (error) {
+					if (completion) {
+						[self.callbackQueue addOperationWithBlock:^{
+							completion(error);
+						}];
+					}
+				} else {
+					// Check again if we have new pending ops while we synced, and repeat as needed
```
shrishrirang: This comment is not valid anymore as we are not calling pullWithQueryInternal again. Can you update it?

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,55 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
+		// Before we can pull from the remote, we need to make sure out table doesn't having pending operations
+		NSArray *tableOps = [self.operationQueue getOperationsForTable:query.table.name item:nil];
+		if (tableOps.count > 0) {
+			NSOperation *push = [self pushWithCompletion:^(NSError *error) {
+				// For now we just abort the pull if the push failed to complete successfully
+				// Long term we can be smarter and check if our table succeeded
+				if (error) {
+					if (completion) {
+						[self.callbackQueue addOperationWithBlock:^{
+							completion(error);
```
shrishrirang: If push fails, we won't add the pull operation to the queue. Would that not mean that the pull operation would never finish? Should we not explicitly cancel it?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -2184,6 +2258,7 @@ -(void) testPurgePendingOpsOnDifferentTableSuccess
             
             done = YES;
         }];
+		XCTAssertNotNil(purge);
```
shrishrirang: Minor: Can you fix indentation?

sdk/iOS/src/MSSyncContextInternal.h
```diff
@@ -7,6 +7,10 @@
 #import "MSOperationQueue.h"
 #import "MSTable.h"
 
+@class MSQueuePushOperation;
+@class MSQueuePullOperation;
```
shrishrirang: unnecessary now?

sdk/iOS/test/MSSyncTableTests.m
damienpontifex: I had this issue with another PR in the past. I use tab indentation, so lays out correctly in Xcode but GitHub adds the extra indentation when displaying it in the diff (maybe due to the plus symbol next to it)

sdk/iOS/src/MSSyncContext.m
```diff
@@ -404,54 +406,56 @@ - (void) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(
...
     dispatch_async(writeOperationQueue, ^{
         // Before we can pull from the remote, we need to make sure out table doesn't having pending operations
         NSArray *tableOps = [self.operationQueue getOperationsForTable:query.table.name item:nil];
         if (tableOps.count > 0) {
-            [self pushWithCompletion:^(NSError *error) {
+            NSOperation *push = [self pushWithCompletion:^(NSError *error) {
                 // For now we just abort the pull if the push failed to complete successfully
                 // Long term we can be smarter and check if our table succeeded
                 if (error) {
+                    [pull cancel];
```
phvannor: Are you sure calling cancel is enough?  Doesn't that just set the isCancelled flag?  The op itself wouldn't set isFinished until it goes to run.  What flag is the dependent Op observing?

shrishrirang: Cancelling the NSOperation would mark it finished, so the observing task should be fine.

damienpontifex: The operation wouldn't have even started at this point, so I don't think we need to do anything else. The completion block will be called if the user has provided one to notify them of the failure. 

I could do some testing though with the returned pull operation as a dependency of another operation. If the execution flows through to the waiting operation when this is cancelled, should all be fine.

phvannor: Can you link to docs?  Everything I see online, says that the behavior of cancel isn't different if the operation is not running.  So if you want to trigger a dependent task, you are responsible for setting the isFinished flag as well.

damienpontifex: Running through a unit test, you were right @phvannor that without explicitly setting the finished status, the dependent tasks wouldn't be completed.

Latest commit with changes to access internal method for completion plus unit test for this case added.

shrishrirang: Thanks for adding UT to test this path.
Btw, should we not simply add the pull NSOperation to the queue immediately after making it dependent on the push NSOperation (instead of waiting for the push to complete successfully)? That way you won't need to call 'completeOperation' explicitly. Just calling cancel ( [push cancel] ) will be enough.

phvannor: Sadly, then the pull possibly already be running, we'd actually need the push op to expose its error state, so the pull op could look at its dependencies & verify they are were ok before running.  Probably worth looking into, but not sure on effort to do so vs what we have now.

shrishrirang: Right!

PR #714, opened 2015-05-26
Push/Pull cancellability UTs

shrishrirang:
- Added UTs to test cancellability of push and pull operations at various execution points
- Removed UT testPullResultingInErrorAllowsDependentOperationsToComplete as completion of push/pull operation is covered by the newly added UTs

### DIFF
--- a/sdk/iOS/test/MSSyncTableTests.m
+++ b/sdk/iOS/test/MSSyncTableTests.m
@@ -1089,15 +1089,15 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
 // - Verifies that the expected number of items are pushed to the server
 -(void) testPushCancellability
 {
-    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:@200], 0, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:@200], 0, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:@200], 1, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:@200], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:200], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:200], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:200], 1, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:200], 0, "Push cancellation failed");
 
-    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:@500], 0, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:@500], 0, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:@500], 0, "Push cancellation failed");
-    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:@500], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:500], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:500], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:500], 0, "Push cancellation failed");
+    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:500], 0, "Push cancellation failed");
 }
 
 // Performs push and cancels it at the specified cancellation point.
@@ -1105,10 +1105,10 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
 // - Verifies that cancelling a push finishes the push NSOperation regardless of the cancellation point
 // - Verifies that the expected number of items are pushed to the server
 // - Returns the number of items pushed to the server.
--(int) performPushWithCancellationPoint:(int) requestedCancellationPoint serverResponseCode:(NSNumber *)serverResponseCode
+-(int) performPushWithCancellationPoint:(int) requestedCancellationPoint serverResponseCode:(NSInteger)serverResponseCode
 {
     // Initialization
-    MSMultiRequestTestFilter *filter = [MSMultiRequestTestFilter testFilterWithStatusCodes:@[serverResponseCode] data:@[@"{\"id\": \"id1\", \"text\":\"server\"}"] appendEmptyRequest:YES];
+    MSTestFilter *filter = [MSTestFilter testFilterWithStatusCode:serverResponseCode data:@"{\"id\": \"id1\", \"text\":\"server\"}"];
     MSClient *filteredClient = [client clientWithFilter:filter];
     MSSyncTable *todoTable = [filteredClient syncTableWithName:TodoTableNoVersion];
 
@@ -1119,8 +1119,8 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
 
     __block NSOperation *pushOperation;
 
-    // Define filters
-    ((MSTestFilter *)filter.testFilters[0]).onInspectRequest = ^(NSURLRequest *request) {
+    // Define the filter
+    filter.onInspectRequest = ^(NSURLRequest *request) {
 
         // Cancellation point 0
         if (requestedCancellationPoint == 0) {
@@ -1130,7 +1130,7 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
         return request;
     };
     
-    ((MSTestFilter *)filter.testFilters[0]).onInspectResponseData = ^(NSURLRequest *request, NSData *data) {
+    filter.onInspectResponseData = ^(NSURLRequest *request, NSData *data) {
 
         // Cancellation point 1
         if (requestedCancellationPoint == 1) {
--- a/sdk/iOS/test/MSSyncTableTests.m
+++ b/sdk/iOS/test/MSSyncTableTests.m
@@ -1087,11 +1087,35 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
 // - Verifies that a push following a cancelled push operation works as expected
 // - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
 // - Verifies that the expected number of items are pushed to the server
--(void) testPushCancellability_successfulpush
+-(void) testPushCancellability_SuccessfulPush_CancelBeforeSendingPushRequest
 {
     XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:200], 0, "Push cancellation failed at cancellation point 0");
+}
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_SuccessfulPush_CancelWhileProcessingPushreponse
+{
     XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:200], 0, "Push cancellation failed at cancellation point 1");
+}
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_SuccessfulPush_CancelFromCompletionBlock
+{
     XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:200], 1, "Push cancellation failed at cancellation point 2");
+}
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_SuccessfulPush_CancelRightAfterPushWithCompletionReturns
+{
     XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:200], 0, "Push cancellation failed at cancellation point 3");
 }
 
@@ -1099,11 +1123,36 @@ static NSString *const SyncContextQueueName = @"Sync Context: Operation Callback
 // - Verifies that a push following a cancelled push operation works as expected
 // - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
 // - Verifies that the expected number of items are pushed to the server
--(void) testPushCancellability_unsuccessfulpush
+-(void) testPushCancellability_UnsuccessfulPush_CancelBeforeSendingPushRequest
 {
     XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:500], 0, "Push cancellation failed at cancellation point 0");
+}
+
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_UnsuccessfulPush_CancelWhileProcessingPushreponse
+{
     XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:500], 0, "Push cancellation failed at cancellation point 1");
+}
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_UnsuccessfulPush_CancelFromCompletionBlock
+{
     XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:500], 0, "Push cancellation failed at cancellation point 2");
+}
+
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability_UnsuccessfulPush_CancelRightAfterPushWithCompletionReturns
+{
     XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:500], 0, "Push cancellation failed at cancellation point 3");
 }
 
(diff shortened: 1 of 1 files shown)


sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1878,54 +2218,6 @@ -(void) testIncrementalPullWithOrderByFails
     XCTAssertEqual(1, filter.actualRequests.count);
 }
 
-- (void)testPullResultingInErrorAllowsDependentOperationsToComplete
```
phvannor: Why is this test removed?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
     XCTAssertEqual(filteredClient.syncContext.pendingOperationsCount, 3);
 }
 
+// Tests cancellation of the push operation at various stages of a push workflow.
+// - Verifies that a push following a cancelled push operation works as expected
+// - Verifies that cancelling a push operation finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+-(void) testPushCancellability
```
phvannor: I don't like this style of test for unit tests, it doesn't play nice (in my opinion) with the FW involved.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+// - Returns the number of items pushed to the server.
+-(int) performPushWithCancellationPoint:(int) requestedCancellationPoint serverResponseCode:(NSNumber *)serverResponseCode
+{
+    // Initialization
+    MSMultiRequestTestFilter *filter = [MSMultiRequestTestFilter testFilterWithStatusCodes:@[serverResponseCode] data:@[@"{\"id\": \"id1\", \"text\":\"server\"}"] appendEmptyRequest:YES];
+    MSClient *filteredClient = [client clientWithFilter:filter];
+    MSSyncTable *todoTable = [filteredClient syncTableWithName:TodoTableNoVersion];
+
+    // Start with a clean slate
+    [[todoTable forcePurgeWithCompletion:^(NSError *error) {
```
phvannor: Tests use an in memory store, why did you need to add a force purge here?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+        XCTAssertNil(error, @"error should have been nil.");
+    }] waitUntilFinished];
+
+    __block NSOperation *pushOperation;
+
+    // Define filters
+    ((MSTestFilter *)filter.testFilters[0]).onInspectRequest = ^(NSURLRequest *request) {
+
+        // Cancellation point 0
+        if (requestedCancellationPoint == 0)
```
phvannor: nit: style, always use { }

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    }];
+    
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+    
+    // Perform a push
+    pushOperation = [filteredClient.syncContext pushWithCompletion:^(NSError *error) {
+        
+        // Cancellation point 2
+        if (requestedCancellationPoint == 2)
+            [pushOperation cancel];
```
phvannor: this is cancelling after the op has finished?  Is that where you intended to do this?  Or did you mean to get one for the delegate version of push complete?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+
+// Performs push and cancels it at the specified cancellation point.
+// - Verifies that a push operation following a cancelled push operation works as expected
+// - Verifies that cancelling a push finishes the push NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pushed to the server
+// - Returns the number of items pushed to the server.
+-(int) performPushWithCancellationPoint:(int) requestedCancellationPoint serverResponseCode:(NSNumber *)serverResponseCode
+{
+    // Initialization
+    MSMultiRequestTestFilter *filter = [MSMultiRequestTestFilter testFilterWithStatusCodes:@[serverResponseCode] data:@[@"{\"id\": \"id1\", \"text\":\"server\"}"] appendEmptyRequest:YES];
```
phvannor: Why do you need to use a multirequest filter?  I only counted the one push.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+{
+    // Define a filter that returns an appropriate server response for the requested item
+    MSTestFilter *filter = [MSTestFilter new];
+    MSClient *filteredClient = [client clientWithFilter:filter];
+
+    filter.ignoreNextFilter = YES;
+    filter.onInspectResponseData = ^(NSURLRequest *request, NSData *data) {
+        NSError *error = nil;
+        
+        NSDictionary *item = [filteredClient.serializer itemFromData:request.HTTPBody withOriginalItem:nil ensureDictionary:YES orError:&error];
```
phvannor: This isn't a safe assumption in a generic push method, deletes don't send an item in them.  Ideally push should have a couple short unit tests verifying shape & body, not mixed in with cancel tests.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+
+    return synchronizedItemCount;
+}
+
+// Performs a push operation and returns the count of all client items pushed so far
+-(int) performVanillaPush
+{
+    // Define a filter that returns an appropriate server response for the requested item
+    MSTestFilter *filter = [MSTestFilter new];
+    MSClient *filteredClient = [client clientWithFilter:filter];
```
phvannor: This ends up trumping the filter setup by the actual test case, I'm a little lost as to the intent here.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    }];
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+    
+    return [self synchronizedItemCount];
+}
+
+// Returns the number of synchronized items, i.e. items that are present on the server as well as the client.
+// An item is synchronized if it was successfully pushed to the server from the client OR
+// if it was successfully pulled from the server to the client.
+-(int) synchronizedItemCount
```
phvannor: Why do you need this?  The test offline store, already tracks a number of simple metrics.  What are we gaining by adding a read here as well?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    
+    int synchronizedItemCount = [self synchronizedItemCount];
+    
+    XCTAssertEqual([self performVanillaPull], 3, "Expected all server items to have been pulled to the client by now");
+    
+    return synchronizedItemCount;
+}
+
+// Performs a pull operation and returns the count of all the server items pulled so far
+-(int) performVanillaPull
```
phvannor: We really care more about the incremental pulls than the defaults.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    MSQuery *query = [[MSQuery alloc] initWithSyncTable:todoTable];
+
+    __block bool isFirstPullRequest = true;
+    filter.ignoreNextFilter = YES;
+    filter.onInspectResponseData = ^(NSURLRequest *request, NSData *data) {
+
+        NSDictionary *item = [filteredClient.serializer itemFromData:request.HTTPBody withOriginalItem:nil ensureDictionary:YES orError:nil];
+        
+        NSString *responseString;
+        if (item != nil) // This means it is a push request
```
phvannor: This is why the multirequest filter exists, to let the filter handle returning responses/etc.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: This test verifies that the pull NSOperation is in finished state if the push fails. There is just one of many cases (error at different points in the workflow and not just during push, no error case, etc) in which we need to test for successful completion of the NSOperation. All this is being already covered as part of the 2 push and pull UTs added.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: I agree in general it is non-ideal to couple multiple tests. However, in this case, we are essentially just testing cancel. The 3 verification bullets listed are all attributes of a clean cancel and not really different test cases.

phvannor: I'd much rather see each case broken out, and named in a readable manner.  I'd prefer no shared code involving asserts, but at a minimum it should be immediately obv what one failed & simple to replay a single case.

shrishrirang: OK, I am planning of breaking down the push test into the following (and similar for pull). This way there are 2 tests:
- testpush_200: a successful push that gets cancelled at various possible points during its execution
- testpush_500: an unsuccessful push that gets cancelled at various possible points during its execution

Shared code will continue to have assertions that should hold true irrespective of the cancellation point and error codes.

-(void) testPushCancellability_200
{
    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:@200], 0, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:@200], 0, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:@200], 1, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:@200], 0, "Push cancellation failed");
}

-(void) testPushCancellability_500
{
    XCTAssertEqual([self performPushWithCancellationPoint:0 serverResponseCode:@500], 0, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:1 serverResponseCode:@500], 0, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:2 serverResponseCode:@500], 0, "Push cancellation failed");
    XCTAssertEqual([self performPushWithCancellationPoint:3 serverResponseCode:@500], 0, "Push cancellation failed");
}

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: The in-memory store will be initialized before testPushCancellability is called, but we are calling performPushWithCancellationPoint multiple times from within it.

sdk/iOS/test/MSSyncTableTests.m
phvannor: Its still hard from that to say why it failed, I'd prefer at the least to have the assert describe the failure reason.  I'd actually just like each assert to be its own test function really for better interplay with the bots & reporting.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: The completion callback is called asynchronously before the operation is completed, so the operation can be in progress when we cancel.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: In an earlier iteration two items were inserted, not anymore. Fixed it.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: This is not a generic push. The idea is to check that a cancelled push does not put the internal data structures in a weird state such that no future push operations go through. This method simply performs a very basic (but not generic) push and ensures that it completes fine. If the method name appears confusing, I can change that.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Yes, that's intentional. This push is performed without cancellation after a prior push operation was cancelled. The idea is to see if this completes successfully to verify that the cancelled push operation did not mess up internal data structures such that no future push operations go through.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: This method is used to count the number of items that have been pushed (or pulled) at any point of time (as part of all the past push or pull operations).

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Not sure I understand your comment. We are not testing incremental pull here. We are just making sure future pull operations continue to work fine after cancelling a pull. This method just performs a very basic pull.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: This method is called more than once and the responses expected are a function of when cancel was performed. We don't want a statically configured filter.

sdk/iOS/test/MSSyncTableTests.m
phvannor: Yes, this method is confusing to me.  Its defintely not a vanilla push then, and its assuming alot of internal state making it only work for specific test cases that is unclear why I would use it.

sdk/iOS/test/MSSyncTableTests.m
phvannor: Its an odd mechanic for a helper to override the parent's filter, esp given push is really dependent on the parent's actions up to now.

sdk/iOS/test/MSSyncTableTests.m
phvannor: The helper gives you base stats on upserted items, is this count going to be different?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    __block int numSynchronizedItems = 0;
+
+    XCTestExpectation *readExpectation = [self expectationWithDescription:@"read expectation"];
+
+    // Read from the table
+    [client.syncContext readWithQuery:query completion:^(MSQueryResult *result, NSError *error) {
+        
+        // Count the number of synchronized items
+        for (NSDictionary* item in result.items) {
+            if ([item[@"text"] isEqualToString:@"server"]) {
```
phvannor: This should just be made part of the query, no reason to loop over results when you could let the store do the filtering.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,348 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+        [readExpectation fulfill];
+    }];
+    
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+    
+    return numSynchronizedItems;
+}
+
+// Verify that the specified NSOperation has finished
+-(void) verifyFinished:(NSOperation *)operation
```
phvannor: Cant this just use a completionblock on the NSOperation itself to verify its complete?

sdk/iOS/test/MSSyncTableTests.m
phvannor: Just that we don't expect ppl to use the basic pull, only inc pulls in the wild.  So unit test coverage can focus on the inc pull codepaths if we don't cover both sets in UTs

sdk/iOS/test/MSSyncTableTests.m
phvannor: Which is why I don't understand overriding the filter that is setup by the base unit test, if the responses adjust dynamically anyways its unclear what we are testing.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Ok, I'll make the assert message more detailed.
I'm avoiding splitting the 4 calls in into methods of their own because the only difference in them would be the cancellation points and they are kinda arbitrary points in the execution workflow. (Same for pull).

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Done.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: The NSOperation returned by the push/pull methods could potentially be already finished. We may not get a chance to add the completion block to the NSOp before it is finished.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Do you mean the upsertedItems, upsertedCalls, etc metrics? They track the upserted items in all tables (the table we are inserting in, operations table, errors table) not just the table we are inserting in. We just care about the final state of the user table.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Updated the method name.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: The caller will now decide which client is to be used. The method now accepts the client as a param.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: The UT is verifying that the expected items are inserted in the local store after the cancelled push is completed. I think it should suffice to just rely on the fact that the cancelled pull leaves the data store in the correct state.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: There is a difference in what the first filter does and what the filter in performVanillaPush (method renamed now) does.
The first filter expects a push request for the inserted item followed by a pull request and is statically configured to respond accordingly.
The second filter may get variable number of push requests depending on when the earlier pull was cancelled. This filter looks at the requests and responds accordingly dynamically. This is the reason for overriding the first filter.

sdk/iOS/test/MSSyncTableTests.m
phvannor: So can't we just call [operation waitUntilFinished]?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,380 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
 
 #pragma mark Pull Tests
 
+// Tests cancellation of the pull operation at various stages of the pull workflow
+// - Verifies that a pull operation following a cancelled pull operation works as expected
+// - Verifies that cancelling a pull operation finishes the pull NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pulled from the server
+-(void) testPullCancellability_successfulpush_successfulpull
+{
+    XCTAssertEqual([self performPullWithCancellationPoint:0 serverPushResponseCode:@200 serverPullResponseCode:@200], 1, "Pull cancellation failed at cancellation point 0");
```
phvannor: I would like better comments on each point, so its clear from reading the code what each "point" is testing.  Its unclear now to just have code say == 0, 1, etc and later on we dont know why these points where picked.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1082,9 +1083,380 @@ -(void) testPushNetworkTimeoutWithErroredOps
...
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+}
 
 #pragma mark Pull Tests
 
+// Tests cancellation of the pull operation at various stages of the pull workflow
+// - Verifies that a pull operation following a cancelled pull operation works as expected
+// - Verifies that cancelling a pull operation finishes the pull NSOperation regardless of the cancellation point
+// - Verifies that the expected number of items are pulled from the server
+-(void) testPullCancellability_successfulpush_successfulpull
```
phvannor: I would still prefer each cancel point be its own test method.  Its unclear what is gained by combining them like so. Now you are dealing with previous test state/etc that would just be avoided by breaking them all up.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: we want to verify that it indeed finishes. waituntilfinish will block forever if the nsop hasn't finished.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Agreed. Done.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Done

shrishrirang:
Ping..

PR #715, opened 2015-05-26
[iOS] nullability annotations on all public headers

damienpontifex:
Continued nullability annotations on all public headers for iOS SDK after #704


azurecla:

TTYL, AZPRBOT;

sdk/iOS/src/MSUser.h
```diff
@@ -30,7 +30,7 @@
 
 
 /// The user id of the end user.
-@property (nonatomic, copy, readonly)   NSString *userId;
+@property (nonatomic, copy, readonly, nonnull)   NSString *userId;
```
phvannor: this can be null, all log out does is set this to null iirc.

sdk/iOS/src/MSUser.h
```diff
@@ -40,7 +40,7 @@
 /// requests made to the Microsoft Azure Mobile Service, allowing the client to
 /// perform all actions on the Microsoft Azure Mobile Service that require
 /// authenticated-user level permissions.
-@property (nonatomic, copy)         NSString *mobileServiceAuthenticationToken;
+@property (nonatomic, copy, nonnull)         NSString *mobileServiceAuthenticationToken;
```
phvannor: same here, I believe null is allowed.

sdk/iOS/src/MSTableOperationError.h
```diff
@@ -59,25 +59,25 @@
 
 /// Removes the pending operation so it will not be tried again the next time push is called. In addition,
 /// updates the local store state
-- (void) cancelOperationAndUpdateItem:(NSDictionary *)item completion:(MSSyncBlock)completion;
+- (void) cancelOperationAndUpdateItem:(nonnull NSDictionary *)item completion:(nullable MSSyncBlock)completion;
 
 /// Removes the pending operation so it will not be tried again the next time push is called. In addition,
 /// removes the item associated with the operation from the local store
-- (void) cancelOperationAndDiscardItemWithCompletion:(MSSyncBlock)completion;
+- (void) cancelOperationAndDiscardItemWithCompletion:(nonnull MSSyncBlock)completion;
```
phvannor: blocks should be nullable

sdk/iOS/src/MSTableOperationError.h
```diff
@@ -19,27 +19,27 @@
...
 @property (nonatomic, readonly) MSTableOperationTypes operation;
 
 /// The id of the item to the operation ran for
-@property (nonatomic, readonly, copy) NSString *itemId;
+@property (nonatomic, readonly, copy, nullable) NSString *itemId;
 
 /// The full item being sent to the server, this item may not always be present for all
 /// operations
-@property (nonatomic, readonly, copy) NSDictionary *item;
+@property (nonatomic, readonly, copy, nonnull) NSDictionary *item;
```
phvannor: nullable.

sdk/iOS/src/MSTableOperation.h
```diff
@@ -28,13 +28,13 @@ typedef NS_OPTIONS(NSUInteger, MSTableOperationTypes) {
 @property (nonatomic, readonly) MSTableOperationTypes type;
 
 /// The name of the table associated with the item
-@property (nonatomic, copy, readonly) NSString *tableName;
+@property (nonatomic, copy, readonly, nonnull) NSString *tableName;
 
 /// The Id of the item the operation should run on.
-@property (nonatomic, copy, readonly) NSString *itemId;
+@property (nonatomic, copy, readonly, nullable) NSString *itemId;
```
phvannor: why is this one nullable, I'd have assumed all ops have itemIds and its item that is the nullable one.

sdk/iOS/src/MSTable.h
```diff
@@ -162,12 +162,12 @@ extern NSString *const MSSystemColumnDeleted;
...
 /// instance provides more flexibilty when querying a table than the table
 /// read* methods.
--(MSQuery *)query;
+-(nonnull MSQuery *)query;
 
 /// Returns an *MSQuery* instance associated with the table that uses
 /// the given predicate. An *MSQuery* instance provides more flexibilty when
 /// querying a table than the table read* methods.
--(MSQuery *)queryWithPredicate:(NSPredicate *)predicate;
+-(nonnull MSQuery *)queryWithPredicate:(nullable NSPredicate *)predicate;
```
phvannor: seems like this one could be nonnull, why use it if you aren't passing a predicate.

sdk/iOS/src/MSTable.h
```diff
@@ -129,31 +129,31 @@ extern NSString *const MSSystemColumnDeleted;
...
 /// from the table. The Microsoft Azure Mobile Service will apply a default
 /// limit to the number of items returned.
--(void)readWithCompletion:(MSReadQueryBlock)completion;
+-(void)readWithCompletion:(nullable MSReadQueryBlock)completion;
 
 /// Sends a request to the Microsoft Azure Mobile Service to return all items
 /// from the table that meet the conditions of the given predicate.
--(void)readWithPredicate:(NSPredicate *) predicate
-      completion:(MSReadQueryBlock)completion;
+-(void)readWithPredicate:(nullable NSPredicate *) predicate
```
phvannor: nonnull here?

sdk/iOS/src/MSTable.h
```diff
@@ -129,31 +129,31 @@ extern NSString *const MSSystemColumnDeleted;
...
+       parameters:(nullable NSDictionary *)parameters
+       completion:(nullable MSItemBlock)completion;
 
 /// Sends a request to the Microsoft Azure Mobile Service to return all items
 /// fromm the table that meet the conditions of the given query.
 /// You can also use a URI in place of queryString to fetch results from a URI e.g.
 /// result.nextLink gives you URI to next page of results for a query that you can pass here.
--(void)readWithQueryString:(NSString *)queryString
-                completion:(MSReadQueryBlock)completion;
+-(void)readWithQueryString:(nullable NSString *)queryString
```
phvannor: we allow it, but seems that if you dont want a QS, you'd just use teh readWithCompletion version, so this could be flagged nonnull?

sdk/iOS/src/MSSyncTable.h
```diff
@@ -98,17 +98,17 @@
...
 /// the server. If a pending request for this table fails, the pull will be cancelled
--(void)pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(MSSyncBlock)completion;
+-(void)pullWithQuery:(nonnull MSQuery *)query queryId:(nullable NSString *)queryId completion:(nullable MSSyncBlock)completion;
 
 /// Removes all records in the local cache that match the results of the specified query.
 /// If query is nil, all records in the local table will be removed.
 /// Before local data is removed, a check will be made for pending operations on this table. If
 /// any are found the purge will be cancelled and an error returned.
--(void)purgeWithQuery:(MSQuery *)query completion:(MSSyncBlock)completion;
+-(void)purgeWithQuery:(nullable MSQuery *)query completion:(nullable MSSyncBlock)completion;
```
phvannor: nonnull query?

sdk/iOS/src/MSSyncTable.h
```diff
@@ -98,17 +98,17 @@
 /// MSQeury object.
 /// Before a pull is allowed to run, all pending requests on the specified table will be sent to
 /// the server. If a pending request for this table fails, the pull will be cancelled
--(void)pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(MSSyncBlock)completion;
+-(void)pullWithQuery:(nonnull MSQuery *)query queryId:(nullable NSString *)queryId completion:(nullable MSSyncBlock)completion;
```
phvannor: query could be nullable here, queryId however shouldn't be.

sdk/iOS/src/MSSyncContextReadResult.h
```diff
@@ -13,7 +13,7 @@
 /// @{
 
 /// Creates a MSSyncContextReadResult with the given items and total count.
-- initWithCount:(NSInteger)count items:(NSArray *)items;
+- (nonnull instancetype)initWithCount:(NSInteger)count items:(nonnull NSArray *)items;
```
phvannor: hmm, so if I get 0 records back, array is an empty array not nil today?

sdk/iOS/src/MSSyncContext.h
```diff
@@ -54,19 +54,19 @@
...
--(NSDictionary *) readTable:(NSString *)table withItemId:(NSString *)itemId orError:(NSError **)error;
+-(nullable NSDictionary *) readTable:(nonnull NSString *)table withItemId:(nonnull NSString *)itemId orError:(NSError * __nullable * __nullable)error;
 
 /// Should insert/update the given item in the local store as appropriate
--(BOOL) upsertItems:(NSArray *)item table:(NSString *)table orError:(NSError **)error;
+-(BOOL) upsertItems:(nullable NSArray *)item table:(nonnull NSString *)table orError:(NSError * __nullable * __nullable)error;
 
 /// Should remove the provided item from the local store
--(BOOL) deleteItemsWithIds:(NSArray *)items table:(NSString *)table orError:(NSError **)error;
+-(BOOL) deleteItemsWithIds:(nullable NSArray *)items table:(nonnull NSString *)table orError:(NSError * __nullable * __nullable)error;
```
phvannor: items should be nonnull I'd assume.

sdk/iOS/src/MSSyncContext.h
```diff
@@ -54,19 +54,19 @@
...
 
 /// Should remove all entries from the specified table in the local store
--(BOOL) deleteUsingQuery:(MSQuery *)query orError:(NSError **)error;
+-(BOOL) deleteUsingQuery:(nullable MSQuery *)query orError:(NSError * __nullable * __nullable)error;
```
phvannor: What happens if I don't pass a query?  Delete all items?  I guess nullable is ok, I'd have to look up what we do.

sdk/iOS/src/MSUser.h
damienpontifex: I though logout just set currentUser to nil in [MSClient](https://github.com/Azure/azure-mobile-services/blob/MobileApp/sdk/iOS/src/MSClient.m#L213-L216)

sdk/iOS/src/MSUser.h
damienpontifex: In the serialisation of the user in [MSLoginSerializer](https://github.com/Azure/azure-mobile-services/blob/MobileApp/sdk/iOS/src/MSLoginSerializer.m#L92-L103) the user isn't construct without the userId and authToken being NSString instances. Hence why I thought they should be nonnullable. Also as with the logout, it just sets the currentUser to nil rather than specific properties

sdk/iOS/src/MSTable.h
damienpontifex: I had this as nullable as in the other constructor `-[MSQuery initWithTable:]` the predicate parameter is passed as nil. Hence assumed the parameter was nullable further up the tree. But in the context of this one call, you would just use the `-[MSTable query]` method if you had a nil predicate.

sdk/iOS/src/MSTable.h
damienpontifex: This was the same thought as queryWithPredicate being potentially nullable to read.

sdk/iOS/src/MSSyncTable.h
damienpontifex: In the implementation, you have a check for query == nil here https://github.com/Azure/azure-mobile-services/blob/MobileApp/sdk/iOS/src/MSSyncTable.m#L70-L74

Hence the nullable annotation.

sdk/iOS/src/MSSyncTable.h
damienpontifex: For the queryId when passed to `-[MSSyncContext pullWithQuery:queryId:completion:]` you do a heap of checks for `if (queryId)` hence I assumed it could be nullable.

sdk/iOS/src/MSSyncContextReadResult.h
damienpontifex: The only use I found was here: https://github.com/Azure/azure-mobile-services/blob/MobileApp/sdk/iOS/src/MSCoreDataStore.m#L201-L306

For this, the results array is allocated and assigned inside the performBlockAndWait. Hence I assumed it would be nonnull and just empty if there were zero results.

sdk/iOS/src/MSSyncContext.h
damienpontifex: Well the query is used to construct the NSFetchRequest. If it or all it's properties are nil then the predicate, sort etc of the fetch request will just be nil and hence I assumed the query could be nullable.

sdk/iOS/src/MSUser.h
```diff
@@ -18,7 +18,7 @@
 ///@{
 
 /// Initializes an *MSUser* instance with the given user id.
--(id)initWithUserId:(NSString *)userId;
+-(nonnull instancetype)initWithUserId:(nonnull NSString *)userId;
```
phvannor: We have a unit test: -(void) testInitWithUserIdAllowsNilUserId, that says you can use nil here.  To me, since the auth token is the only important data it makes sense to allow no user Id (Its been awhile, but I don't remember caching my userIds in the past when building auth out)

sdk/iOS/src/MSUser.h
phvannor: User doesn't have a constructor taking both?  The client example shows code as so:

  self.todoService.client.currentUser = [[MSUser alloc] initWithUserId:userid];
  self.todoService.client.currentUser.mobileServiceAuthenticationToken = [SSKeychain passwordForService:@"AzureMobileServiceTutorial" account:userid];

So its possible for the token to be null in a valid flow

sdk/iOS/src/MSSyncTable.h
phvannor: oh yes, you are correct, guess I was a little tired yesterday.

sdk/iOS/src/MSSyncContextReadResult.h
phvannor: ok, fine for now, may need revisited long term.

sdk/iOS/src/MSSyncContext.h
phvannor: I guess it will work, but not the intent.  Since any 3rd party could conform to this, I think it feels cleaner to mark it as non null so there is not differing behavior long term if its passed as nil.

damienpontifex:
@phvannor was there anymore changes/feedback on this?

phvannor:
@damienpontifex I was out the last couple work days, I'll take a final pass today/tomorrow to refresh my memory, I think it was OK now.

damienpontifex:
No worries, was just doing my Monday thing to review open items

sdk/iOS/src/MSTableOperation.h
```diff
@@ -28,13 +28,13 @@ typedef NS_OPTIONS(NSUInteger, MSTableOperationTypes) {
...
-@property (nonatomic, copy, readonly) NSString *tableName;
+@property (nonatomic, copy, readonly, nonnull) NSString *tableName;
 
 /// The Id of the item the operation should run on.
-@property (nonatomic, copy, readonly) NSString *itemId;
+@property (nonatomic, copy, readonly, nonnull) NSString *itemId;
 
 /// The item that will be sent to the server when execute is called.
-@property (nonatomic, strong) NSDictionary *item;
+@property (nonatomic, strong, nonnull) NSDictionary *item;
```
phvannor: This one should be nullable, the internal constructor can make one w/o an item, so it is not 100% always around.

phvannor:
:+1:

phvannor:
Looks like it can't merge, can you resolve any conflicts so I can take this?

damienpontifex:
Merged changes. Added nullable annotations to the returned operations we added from #709

PR #723, opened 2015-06-11
[iOS] Add support for configuring page size of a pull operation

shrishrirang:


sdk/iOS/src/MSPullSettings.h
```diff
@@ -0,0 +1,29 @@
...
+///@{
+
+/// Size of pages requested from the server while performing a pull
+@property (nonatomic) NSInteger pageSize;
+
+///@}
+
+#pragma mark * Class methods
+
+///@name Class factory methods
```
phvannor: nit: this is not in the style of apple documentation, should be more action-y, like Creating a MSPullSetting.

sdk/iOS/src/MSPullSettings.h
```diff
@@ -0,0 +1,29 @@
...
+
+/// Class factory method for getting default pull settings
++ (MSPullSettings *)default;
```
phvannor: Why have this?  It doesn't seem to add any value.

sdk/iOS/src/MSPullSettings.m
```diff
@@ -0,0 +1,44 @@
...
+
+#import "MSPullSettings.h"
+
+#pragma mark * MSPullSettings Implementation
+
+@implementation MSPullSettings
+
+NSInteger const defaultPageSize = 50;
+
+@synthesize pageSize = _pageSize;
```
phvannor: nit: remove the _pageSize

sdk/iOS/src/MSPullSettings.m
```diff
@@ -0,0 +1,44 @@
...
+    
+    return self;
+}
+
+#pragma mark * Accessor method(s)
+
+- (void)setPageSize:(NSInteger)pageSize {
+    if (pageSize <= 0) {
+        _pageSize = defaultPageSize;
+    }
```
phvannor: nit: use objC if styling

sdk/iOS/src/MSSyncContextInternal.h
```diff
@@ -27,6 +28,8 @@
 
 -(NSOperation *) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(MSSyncBlock)completion;
 
+-(NSOperation *) pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId pullSettings:(MSPullSettings *)pullSettings completion:(MSSyncBlock)completion;
```
phvannor: settings?  pullSettings seems a lil redundant.  Also, since this is an internal header, just change the existing sig.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
...
+}
+
+-(void) testPullTable_NoPageOptions
+{
+    NSString *pullRequest = [self getPullTableRequestWithPullSettings:nil usePullSettings:NO];
+    
+    XCTAssertEqualObjects(pullRequest, @"https://someUrl/tables/TodoNoVersion?$top=50&__includeDeleted=true&$skip=0&__systemProperties=__deleted", "Incorrect pull request");
+}
+
+-(NSString *) getPullTableRequestWithPullSettings:(MSPullSettings *)pullSettings usePullSettings:(BOOL)shouldUsePullSettings
```
phvannor: nit: this isn't proper objC naming conventions for methods.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
...
+        [[todoTable pullWithQuery:query queryId:nil pullSettings:pullSettings completion:nil] waitUntilFinished];
+    }
+    else {
+        [[todoTable pullWithQuery:query queryId:nil completion:nil] waitUntilFinished];
+    }
+    
+    return pullRequest;
+}
+
+-(void) testPullSyncContext_UsePullSettings_NegativePageSize
```
phvannor: testing sync context internals should probably be done in a synccontext test file, not in MSTable.  Are we already following this pattern?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
                   @"Invalid URL: %@", thirdRequest.URL.absoluteString);
 }
 
+-(void) testPullTable_UsePullSettings_NegativePageSize
```
phvannor: We need at least one test confirming page size is flowing into the inc sync test cases.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
...
+{
+    MSPullSettings *pullSettings = [MSPullSettings new];
+    pullSettings.pageSize = -1;
+    
+    NSString *pullRequest = [self getPullTableRequestWithPullSettings:pullSettings usePullSettings:YES];
+
+    XCTAssertEqualObjects(pullRequest, @"https://someUrl/tables/TodoNoVersion?$top=50&__includeDeleted=true&$skip=0&__systemProperties=__deleted", "Incorrect pull request");
+}
+
+-(void) testPullTable_UsePullSettings_ZeroPageSize
```
phvannor: I think these tests could just be moved to a test class for MSPullSettings and verify just it.  No need to go through the whole pipeline on them.

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
...
+{
+    MSPullSettings *pullSettings = [MSPullSettings new];
+    pullSettings.pageSize = 0;
+    
+    NSString *pullRequest = [self getPullTableRequestWithPullSettings:pullSettings usePullSettings:YES];
+    
+    XCTAssertEqualObjects(pullRequest, @"https://someUrl/tables/TodoNoVersion?$top=50&__includeDeleted=true&$skip=0&__systemProperties=__deleted", "Incorrect pull request");
```
phvannor: you've confirmed in all these tests cases that these sets sort identically on 32 & 64 bit runs?

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1584,6 +1584,142 @@ -(void) testPullWithFetchLimitNotMultipleOfFifty
...
+    
+    if (shouldUsePullSettings) {
+        [[filteredClient.syncContext pullWithQuery:query queryId:nil pullSettings:pullSettings completion:nil] waitUntilFinished];
+    }
+    else {
+        [[filteredClient.syncContext pullWithQuery:query queryId:nil completion:nil] waitUntilFinished];
+    }
+    
+    return pullRequest;
+}
```
phvannor: I dont see a test for the MaxRecords vs PageSize check you updated.

sdk/iOS/src/MSPullSettings.h
shrishrirang: That code is removed in the next iteration

sdk/iOS/src/MSPullSettings.h
shrishrirang: removed

sdk/iOS/src/MSPullSettings.m
shrishrirang: removed synthesized. relying on auto synthesis

sdk/iOS/src/MSPullSettings.m
shrishrirang: done

sdk/iOS/src/MSSyncContextInternal.h
shrishrirang: Settings is too generic. I was planning of calling it sync settings, but not all the pull settings would be shared with push. Later we can have PullSettings and PushSettings derive from SyncSettings if needed.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: yes, all sync context tests are in the same file. Sticking to that.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: added

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: done

sdk/iOS/src/MSSyncContextInternal.h
phvannor: I meant the method name, not the class name.  pullWithQuery:queryId:pullSettings:completion, I think can be just settings: in this context.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: I'm not sure I understand this comment. What is it that you think could potentially be different on 32/64-bit machines?

sdk/iOS/src/MSPullSettings.m
```diff
@@ -0,0 +1,37 @@
...
+
+@implementation MSPullSettings
+
+static NSInteger const _defaultPageSize = 50;
+
+#pragma mark * Initializer method(s)
+
+- (id)init {
+    self = [super init];
+    if (self) {
```
phvannor: usually we would just do a lazy instantiation in the getter.

sdk/iOS/src/MSPullSettings.m
```diff
@@ -0,0 +1,37 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// ----------------------------------------------------------------------------
+
+#import "MSPullSettings.h"
+
+#pragma mark * MSPullSettings Implementation
+
+@implementation MSPullSettings
+
+static NSInteger const _defaultPageSize = 50;
```
phvannor: Consts should not start with an _

sdk/iOS/test/MSSyncTableTests.m
```diff
@@ -1405,6 +1406,54 @@ -(void) testPullWithFetchLimit
...
+    XCTAssertNotEqual(pullSettings.pageSize, MSPullSettings.defaultPageSize, "This test requires the custom page size to be different from the default page size");
+    
+    NSOperation *pull = [todoTable pullWithQuery:query queryId:nil pullSettings:pullSettings completion:^(NSError *error) {
+        XCTAssertNil(error, @"Error found: %@", error.description);
+        done = YES;
+    }];
+    
+    XCTAssertNotNil(pull);
+    
+    XCTAssertTrue([self waitForTest:0.1], @"Test timed out.");
```
phvannor: nit: we should try to flip to using expectations when possible.

sdk/iOS/test/MSSyncTableTests.m
phvannor: Due to the way we build URLs, the NSDictionary ordering can be different between the 2, so the URL may be ?A&B on 32 while be B&A on 64bit devices.  Usually this impacts any query that is using system properties iirc.

sdk/iOS/src/MSPullSettings.h
```diff
@@ -0,0 +1,20 @@
...
+/// Settings that control the pull behavior
+@interface MSPullSettings : NSObject
+
+#pragma mark * Properties
+
+///@name Properties that control pull behavior
+///@{
+
+/// Size of pages requested from the server while performing a pull
+@property (nonatomic) NSInteger pageSize;
```
phvannor: add nonull attribute for Swift.

sdk/iOS/src/MSSyncTable.h
```diff
@@ -96,12 +97,19 @@
...
 /// specified table will be sent to the server. If a pending request for this table fails,
 /// the pull will be cancelled
 -(NSOperation *)pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId completion:(MSSyncBlock)completion;
 
+/// Initiates a request to go to the server and get a set of records matching the specified
+/// MSQuery object.
+/// Before a pull is allowed to run, one operation to send all pending requests on the
+/// specified table will be sent to the server. If a pending request for this table fails,
+/// the pull will be cancelled
+-(NSOperation *)pullWithQuery:(MSQuery *)query queryId:(NSString *)queryId pullSettings:(MSPullSettings *)pullSettings completion:(MSSyncBlock)completion;
```
phvannor: these will need null/nonnull decorators on them, all I believe are __nullable.

sdk/iOS/test/MSSyncTableTests.m
phvannor: also make sure you run the unit tests with both types of devices, if you aren't already as other small changes come up as well.

sdk/iOS/src/MSPullSettings.h
```diff
@@ -0,0 +1,20 @@
...
+
+///@}
+
```
damienpontifex: As per the request on #725 could we add a property here of `@property (nonatomic) BOOL userInitiated`? and then in `init` default it to `YES`

sdk/iOS/src/MSPullSettings.m
damienpontifex: Also make the return type `instancetype` for any init methods

sdk/iOS/src/MSSyncContextInternal.h
shrishrirang: Done

sdk/iOS/src/MSPullSettings.m
shrishrirang: Changed the return type to instancetype. 
Not moving initialization to the getter as it is not really a perf impacting operation. In such a case, I prefer initializing (especially with constant values) in the initializer.

sdk/iOS/src/MSPullSettings.m
shrishrirang: Changed as per apple guidelines

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: waitForTest is not really needed, now that pull returns an NSOp.

sdk/iOS/src/MSSyncTable.h
shrishrirang: Should we just create a work item to review all public APIs in MobileApp?

phvannor: Damien already did them, so mainly new apis need this done now.

sdk/iOS/src/MSPullSettings.h
shrishrirang: It'd be good to make that change (and UTs for it) separately. It might be better to just let the user specify the NSQualityOfService instead of inferring it from the source (user initiated, etc) of the task.

sdk/iOS/test/MSSyncTableTests.m
shrishrirang: Done

sdk/iOS/src/MSPullSettings.m
```diff
@@ -0,0 +1,37 @@
...
+
+#pragma mark * MSPullSettings Implementation
+
+@implementation MSPullSettings
+
+static NSInteger const MSDefaultPageSize = 50;
+
+#pragma mark * Initializer method(s)
+
+- (instancetype)init {
```
phvannor: I feel like this class needs an initWithPageSize constructor for now, with only 1 property it feels awkward to use in the current form.

sdk/iOS/test/MSPullSettingsTests.m
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#pragma mark * PageSize tests
+
+- (void)testPullSettings_NegativePageSize
+{
+    MSPullSettings *pullSettings = [MSPullSettings new];
+    pullSettings.pageSize = -1;
+
+    XCTAssertEqual(pullSettings.pageSize, MSPullSettings.defaultPageSize, "Incorrect page size");
```
phvannor: These strings don't have to be valid ObjC syntax?  So maybe its a nit, but I'd like @"Incorrect page size"

sdk/iOS/src/MSPullSettings.m
shrishrirang: Yes, it sure is a bit awkward, but as this class is just a container for miscellaneous settings it will be difficult to have initializers for all properties as the property list grows. At that point it will not be possible to remove initWithPageSize (breaking change) and at the same time having an initializer for page size but not for others would seem weird. What do you think?

phvannor: I think we'd only have constructors for the commonly used setups (page size being one of them)  I'd be fine with not having one that includes the QoS param when that is added.

sdk/iOS/src/MSPullSettings.h
phvannor: Lets do it in a separate log, I'm not sure we'd want the default to be YES on this either.

sdk/iOS/src/MSSyncTable.h
shrishrirang: I can see only MSClient.h being annotated in MobileApp. Am I missing something?

sdk/iOS/src/MSPullSettings.m
shrishrirang: done

sdk/iOS/test/MSPullSettingsTests.m
shrishrirang: Looks like both work (verified that), but from the documentation it is clear it expects NSString. Fixed.

sdk/iOS/src/MSSyncTable.h
phvannor: https://github.com/Azure/azure-mobile-services/commit/d896042264e67e5e7e57a037910421510dc10223#diff-d7d786ab102158791637bc78f2069ce7

shrishrirang: Plan to send another iteration. Do not merge this yet.

shrishrirang: Let's add the annotations in the MobileApp branch. Partially annotating a header file results in warnings. I don't want to annotate the entire header file in dev branch to avoid potential merge conflicts.

shrishrirang:
Merging of the latest changes from dev is in progress. Please don't accept the PR.

phvannor:
if the PR can me merged, you shoudln't need to merge dev into it as you have no conflicts.

shrishrirang:
I had pushed an incompletely merged branch. Just didn't want anyone to accept the PR before I fixed the build break.
The build break has been fixed now. Thanks!

Issue #718, opened 2015-05-29
Elements in the OperationErrorList table cause crash when getting sync table on Android SDK 2.0.2


jonsheller:
During poor Internet connectivity, elements are inserted in the OperationErrorList table. During the next app startup when these elements are loaded, the app crashes with the exception "java.lang.NumberFormatException" because the operationkind stored in the table is stored as a stringified float, and the code that loads it expects an integer.

Example of an item stored in the table:

{"statuscode":0.0,
"errormessage":"com.microsoft.windowsazure.mobileservices.MobileServiceException: There was an error executing the request",
"id":"fe4146fc-c377-47d7-a5aa-68ae69415f9c",
"serveritem":null,
"operationkind":"1.0",
"clientitem":
{
//our application data,
"id":"18cdc0bf-b231-4bf1-83f4-bcc146a699f3"},
"serverresponse":null,
"__createdat":"2015-05-29T11:39:22.374Z",
"tablename":"ourtablename",
"operationId":"0ba47125-d39a-4ba0-9be6-ae6d4cc1a131",
"itemid":"18cdc0bf-b231-4bf1-83f4-bcc146a699f3"}

Note that the stored operationkind is 1.0

Then when loading this item during the next app startup, OperationErrorList.deserialize calls

int operationKind = element.get("operationkind").getAsInt();

But getAsInt assumes that the  value can be parsed as an integer, which it cannot because it contains a decimal point, and at this point parseInt throws the aforementioned NumberFormatException.

jonsheller:
Note that "Application startup" is because it's during application startup that I connect to Azure Mobile Services, so that's the point that the operations are deserialized; it's not that Azure is hooking application lifecycle events by itself.

masojus:
Has anybody had a chance to take a look at this issue? It's very severe, and is a surprisingly easy scenario to hit. During our testing in places with shoddy cell signal and/or wifi, this happens consistently. What this means is that while using the app, if one generates operations in the OperationErrorList table, the app goes into an infinite crash cycle, crashing on boot every time we load up the local store and try to sync.

[...]

PR #731, opened 2015-06-22
Fix issue #718

marianosz:
Fix issue #718


sdk/Javascript/src/External/esprima
```diff
@@ -1 +1 @@
-Subproject commit f5cd34b2af9e56be7c12c6149bbdad672505c241
+Subproject commit 1bfe632f5b207eb57c22b4ee2c10906ba755a83d
```
soninaren: subproject should not be changed, please exclude this

Issue #720, opened 2015-06-04
Error in generated query from pull during request


icsd:
The following request is generated from the Azure Mobile Services SDK  2.0.2 for Android, during pull. The results of the response are not correct because of the OR operator. It seems that a couple parentheses is missing.

GET /tables/groups?$filter= MY_QUERY%20
and%20__updatedAt%20gt%20(datetimeoffset'2015-05-29T13:25:07.312Z')%20
and%20(id%20ne%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))%20
and%20(id%20ne%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))%20
or%20(__updatedAt%20ge%20(datetimeoffset'2015-05-29T13:25:07.312Z')%20
and%20id%20gt%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))
&$top=50&$orderby=__updatedAt%20asc,id%20asc&__includeDeleted=true&__systemproperties=__updatedAt,__version,__deleted HTTP/1.1

I believe it should be like this:
I have added the extra parentheses with bold

GET /tables/groups?$filter= MY_QUERY%20
and%20  **(**  __updatedAt%20gt%20(datetimeoffset'2015-05-29T13:25:07.312Z')%20
and%20(id%20ne%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))%20
and%20(id%20ne%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))%20
or%20(__updatedAt%20ge%20(datetimeoffset'2015-05-29T13:25:07.312Z')%20
and%20id%20gt%20('2E1487FC-D905-46B1-8B2D-FFB8D758553B'))  **)**
&$top=50&$orderby=__updatedAt%20asc,id%20asc&__includeDeleted=true&__systemproperties=__updatedAt,__version,__deleted HTTP/1.1

phvannor:
Query should definitely add parens where shown or else it would break a custom query.

Also, the additional Id not equal to clauses need looked into.  I believe they introduce a bug if the record with that Id has changed in between calls, and I'm not sure why they are present at first glance.

icsd:
Definately the incremental syncing is not working properly. 
Supposing that i want to change paging. Is changing defaultTop in PullStrategy enough? 
If so, i had an issue yesterday where syncing was not working properly and i ended up with (PageSize+1) records.

[...]

PR #732, opened 2015-06-23
Fixed the incremental pull query building logic and the skip reset bug #720

soninaren:
#720 
-- Changed the incremental sync query logic to add parenthesis at appropriate location
-- Removed the redundant and clause in the incremental sync query
-- Fix the skip logic

### DIFF
--- a/sdk/android/test/sdk.testapp/src/main/java/com/microsoft/windowsazure/mobileservices/sdk/testapp/test/MobileServiceSyncTableTests.java
+++ b/sdk/android/test/sdk.testapp/src/main/java/com/microsoft/windowsazure/mobileservices/sdk/testapp/test/MobileServiceSyncTableTests.java
@@ -479,7 +479,7 @@ public class MobileServiceSyncTableTests extends InstrumentationTestCase {
 
         MobileServiceSyncTable<StringIdType> table = client.getSyncTable(StringIdType.class);
 
-        Query query = QueryOperations.tableName(table.getName()).top(2);
+        Query query = QueryOperations.field("String").eq("noMatch");
 
         table.pull(query, "QueryKey").get();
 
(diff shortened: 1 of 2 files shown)


sdk/android/test/sdk.testapp/src/main/java/com/microsoft/windowsazure/mobileservices/sdk/testapp/test/MobileServiceSyncTableTests.java
```diff
@@ -538,6 +537,58 @@ public void testIncrementalPullSaveLastUpdatedAtDate() throws MalformedURLExcept
...
+                        "{\"id\":\"def\",\"String\":\"World\",\"__updatedAt\":\"" + updatedAt1 + "\"}]",
+                "[{\"id\":\"abc\",\"String\":\"Hey2\",\"__updatedAt\":\"" + updatedAt2 + "\"}," +
+                        "{\"id\":\"def\",\"String\":\"World\",\"__updatedAt\":\"" + updatedAt2 + "\"}]"
+        ));
+
+        client.getSyncContext().initialize(store, new SimpleSyncHandler()).get();
+
+        MobileServiceSyncTable<StringIdType> table = client.getSyncTable(StringIdType.class);
+
+        Query query = QueryOperations.tableName(table.getName()).top(2);
```
phvannor: This is not the way .top is supposed to work in pull().

sdk/android/test/sdk.testapp/src/main/java/com/microsoft/windowsazure/mobileservices/sdk/testapp/test/MobileServiceSyncTableTests.java
```diff
@@ -538,6 +537,58 @@ public void testIncrementalPullSaveLastUpdatedAtDate() throws MalformedURLExcept
...
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+        sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
+
+        Date d = new Date();
+        String updatedAt1 = sdf.format(d);
+        d.setTime(d.getTime() + 1000L);
+        String updatedAt2 = sdf.format(d);
+
+        client = client.withFilter(getTestFilter(serviceFilterContainer, false,
+                "{\"count\":\"6\",\"results\":" +
```
phvannor: why are you mocking a count?  A pull shouldn't be asking for inlinecount on the query.  Also, whys it returning 6?  Does the Android filter logic work different than I am used to with managed & iOS?

Issue #699, opened 2015-05-14
[iOS][Offline] Fields with nil value are ignored on push operations


gsach:
When a record is updated locally, and an existing value is replaced with null, this change is not pushed to the server. This causes data loss. A workaround would be to use empty strings instead of null, but this solution doesn't work well for foreign keys.

I inspected the code and found the reason why that issue happens. In the class MSCoreDataStore, the method:

-(id) getRecordForTable:(NSString *)table itemId:(NSString *)itemId asDictionary:(BOOL)asDictionary orError:(NSError **)error

is using NSDictionaryResultType on the fetch request. As a result, the null values are not added to the dictionary which is used to push the changes to the server.

phvannor:
The NSDictionaryResultType was used to support the "selectFields" property of the MSQuery object (via propertiesToFetch).  The readTable:withItemId:orError method appears to be using it for convenience to get a NSDictionary but a simple swap to using tableItemFromManagedObject: method should work here instead.

Also we should consider making tableItemFromManagedObject into an instance method, so MSCoreDataStore could be inherited from and the logic adjusted as needed by the developer better.

phvannor:
This I believe is fixed with Shrirangs PR, closing issue.

PR #733, opened 2015-06-25
[iOS][Offline] Fields with nil value are ignored on push operations

shrishrirang:
This fixes https://github.com/Azure/azure-mobile-services/issues/699
In addition, this also fixes similar issue with readWithQuery.

### DIFF
--- a/sdk/iOS/src/MSCoreDataStore.m
+++ b/sdk/iOS/src/MSCoreDataStore.m
@@ -55,47 +55,40 @@ NSString *const StoreDeleted = @"ms_deleted";
     
     fr.predicate = [NSPredicate predicateWithFormat:@"%K ==[c] %@", MSSystemColumnId, itemId];
     
-    if (asDictionary) {
-        fr.resultType = NSDictionaryResultType;
-    }
-    
     NSArray *results = [self.context executeFetchRequest:fr error:error];
     if (!results || (error && *error)) {
         return nil;
     }
     
-    NSDictionary *item = [results firstObject];
-    if (asDictionary) {
-        // Otherwise this dictionary can't be worked with as user's expect
-        if (item) {
-            return [NSDictionary dictionaryWithDictionary:item];
-        } else {
-            return nil;
-        }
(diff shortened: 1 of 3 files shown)


sdk/iOS/src/MSCoreDataStore.m
```diff
@@ -55,47 +55,40 @@ -(id) getRecordForTable:(NSString *)table itemId:(NSString *)itemId asDictionary
...
+        // For it to work like a regular |NSMutableDictionary| we need to copy the contents into an
+        // |NSMutableDictionary| instance OR into an |NSDictionary| instance and then make a mutable copy.
+        return [NSDictionary dictionaryWithDictionary:result];
     }
     
-    return [results firstObject];
+    return item;
 }
 
 +(NSDictionary *) tableItemFromManagedObject:(NSManagedObject *)object
```
shrishrirang: Leaving this as a class method to avoid a breaking change in the dev branch. We can change this to an instance method once we integrate these changes to MobileApp.

sdk/iOS/src/MSCoreDataStore.m
```diff
@@ -214,7 +207,6 @@ -(MSSyncContextReadResult *)readWithQuery:(MSQuery *)query orError:(NSError *__a
         fr.entity = entity;
         fr.predicate = query.predicate;
         fr.sortDescriptors = query.orderBy;
-        fr.resultType = NSDictionaryResultType;
```
phvannor: Do you know the perf difference in flipping from this to the new model?  Are they roughly eqv?

shrishrirang: I could not find any related documentation and I don't see an easy way to measure it. Another option which I considered, but chose not to use, was to stick to NSDictionaryResultType and check the returned NSDictionary for missing properties. 

However, it might actually make sense to take that approach. We would need to check for missing properties only if the property count of an item does not match the number of properties requested. That way we do extra work (potentially O(n^2) where n == no. of properties) only if an item has one or more null properties.

What do you think?

PR #749, opened 2015-07-16
Update iOS SDK header from v3.0 to v3.1

shrishrirang:


phvannor:
This is still in beta, it should stay 3.0

PR #756, opened 2015-07-22
add doc reference

ysxu:


README.md
```diff
@@ -18,7 +17,19 @@ To get the source code of our SDKs and samples via **git** just type:
     git clone https://github.com/Azure/azure-mobile-services.git
     cd ./azure-mobile-services/
 
-## Change log
+## SDk Documentation references
```
phvannor: K, R ?  Maybe Reference Documentation would be better?

README.md
```diff
@@ -18,7 +17,19 @@ To get the source code of our SDKs and samples via **git** just type:
...
+
+#### Azure App Service Mobile Apps
+* [iOS](http://azure.github.io/azure-mobile-services/ios/v3)
+* [Managed](http://azure.github.io/azure-mobile-services/managed/v2)
```
phvannor: We won't have a managed, android, or JS any time soon here.  Remove or point them at their current sites (if they exist)

README.md
```diff
@@ -18,7 +17,19 @@ To get the source code of our SDKs and samples via **git** just type:
...
+
+#### Mobile Services
+* [iOS](http://azure.github.io/azure-mobile-services/ios/v2)
+* [Managed](http://azure.github.io/azure-mobile-services/managed/v1)
+* [Android](http://azure.github.io/azure-mobile-services/android/v2)
+* [Javascript](http://azure.github.io/azure-mobile-services/javascript/v1)
+
+## SDK Change log
```
phvannor: I dunno why we need SDK really in these headers, the whole readme is on the SDK

phvannor: Also, if we are going to spell out pointers to Apps & services in the readme, we may want to do it for teh change logs as well.  Talk to @lindydonna

README.md
```diff
@@ -33,7 +44,7 @@ as a Nuget package or you can download the source using the instructions above.
 
 ### Prerequisites
 
-The SDK requires Visual Studio 2012 RTM.
+The SDK requires Visual Studio 2015.
```
phvannor: 2013, 15 is a lil early to say we require it.

README.md
```diff
@@ -33,7 +44,7 @@ as a Nuget package or you can download the source using the instructions above.
 
 ### Prerequisites
```
phvannor: The above may want to be revamped to mention Windows Universal & WP 8.1 0r made more generic?

README.md
```diff
@@ -139,21 +150,16 @@ To run the HTML tests:
 2. In the Solution Explorer, select the Microsoft.WindowsAzure.Mobile.WinJS.Test project and right-click to select 'View in Browser'.
 3. The default browser will launch and run the test HTML application. Some tests may fail because due to an 'Unexpected connection failure'. This is because the test is configured to connect to a Mobile Service that does not exist. These failures can be ignored.
 
-## Additional Resources
+## Useful Resources
 
-* [Quickstarts](https://github.com/Azure/azure-mobile-services-quickstarts)
+* [Quickstarts on Github](https://github.com/Azure/azure-mobile-services-quickstarts)
```
phvannor: why call out on Github here?

ysxu: i wanted to differentiate them with the quickstart tutorials up above. but it's not completely necessary.

README.md
ysxu: this is for mobile apps? why would we not have a managed reference? I excluded android and js already.

README.md
ysxu: It might be less confusing for us to differentiate between vnext and mobile services. What do you guys think? Simply giving the versions might be confusing for customers who don't know of App Service @lindydonna

README.md
ysxu: updated.

README.md
ysxu: updated to delete "SDK"

README.md
ysxu: Updated to "
## Managed Windows Client SDK

Our managed portable library for Windows 8, Windows Phone 8, Windows Phone 8.1, and Windows Universal Client SDK makes it incredibly easy to use Mobile Services from your Windows applications."

lindydonna: We shouldn't call it Universal--that term now means UWP on Win10. It looks like we are using the term "Windows Runtime universal C#" in our docs.

README.md
lindydonna: The managed reference docs live on MSDN, and we don't have a way to differentiate versions afaik.

README.md
lindydonna: I have a pending change in the Mobile Apps branch changelog that I still need to fix, which titles things as "Azure Mobile Apps". FYI that's the term we're using in our docs.

The big issue is that the master branch is for Mobile Services, while the MobileApps branch is for vNext. So it's kind of weird to talk about MobileApps while in the master branch. 

Does this readme need to cover both Apps and Services?  If so, then it's not really a readme about the SDK repo anymore.

Issue #751, opened 2015-07-18
try/catch or rethrow needed in background thread (andoird sdk)


rothschild86:
This file https://github.com/Azure/azure-mobile-services/blob/master/sdk/android/src/sdk/src/main/java/com/microsoft/windowsazure/mobileservices/table/MobileServiceTableBase.java

Line 365 should be wrapped:

                    try{
                    	serverEntity = new JsonParser().parse(content).getAsJsonObject();
                    }
                    catch(JsonSyntaxException | IllegalStateException e){
                    	e.printStackTrace();
                    }

If not wrapped and exception is thrown, it never bubbles up to the top. In fact, onCompleted(Throwable throwable) is never even called on table operations.

soninaren:
Fixed in #761

soninaren:
@rothschild86 This is fixed in the referenced Pull Request, will be released with the next android SDK release

PR #761, opened 2015-07-29
Fix Issue #751

marianosz:
Refactor on the method to avoid the parse of the response content if not needed.
Parse method wrapped with exception handling.


soninaren:
#751

PR #775, opened 2015-08-17
First set of changes for JS offline SDK 

shrishrirang:
- This PR contains very basic changes: only insert and lookup
- All operations are being performed on an in-memory store
- Operations are async.
- The next step is to get the same working with a SQLite store, but that will be part of a separate PR.

### DIFF
--- a/sdk/Javascript/src/Microsoft.WindowsAzure.Mobile.JS.csproj
+++ b/sdk/Javascript/src/Microsoft.WindowsAzure.Mobile.JS.csproj
@@ -1,13 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <ToolsPath>$(MSBuildProjectDirectory)\..\tools</ToolsPath>
     <MakePriPath Condition="Exists('$(MSBuildProgramFiles32)\Windows Kits\8.0\bin\x64\makepri.exe')">$(MSBuildProgramFiles32)\Windows Kits\8.0\bin\x64\makepri.exe</MakePriPath>
     <MakePriPath Condition="Exists('..\..\Tools\makepri.exe')">..\..\Tools\makepri.exe</MakePriPath>
+    <TargetPlatformVersion>8.1</TargetPlatformVersion>
+    <MinimumVisualStudioVersion>12</MinimumVisualStudioVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>
+    </TargetFrameworkVersion>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <ProductVersion>8.0.30703</ProductVersion>
--- a/sdk/Javascript/src/Microsoft.WindowsAzure.Mobile.JS.csproj
+++ b/sdk/Javascript/src/Microsoft.WindowsAzure.Mobile.JS.csproj
@@ -186,8 +186,11 @@
     <Content Include="LoginUis\WebAuthBroker.js" />
     <Content Include="MobileServiceClient.js" />
     <Content Include="MobileServiceLogin.js" />
+    <Content Include="MobileServiceMemoryStore.js" />
     <Content Include="MobileServices.intellisense.js" />
     <Content Include="MobileServices.priconfig.xml" />
+    <Content Include="MobileServiceSyncContext.js" />
+    <Content Include="MobileServiceSyncTable.js" />
     <Content Include="MobileServiceTable.js" />
     <Content Include="Platforms\Platform.Web.js" />
     <Content Include="Platforms\Platform.WinJS.js" />
--- a/sdk/Javascript/src/MobileServiceMemoryStore.js
+++ b/sdk/Javascript/src/MobileServiceMemoryStore.js
@@ -1,0 +1,104 @@
+﻿// ----------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// ----------------------------------------------------------------------------
+
+/// <reference path="C:\Program Files (x86)\Microsoft SDKs\Windows\v8.0\ExtensionSDKs\Microsoft.WinJS.1.0\1.0\DesignTime\CommonConfiguration\Neutral\Microsoft.WinJS.1.0\js\base.js" />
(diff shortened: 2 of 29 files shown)


phvannor:
This is going to have to go into its own public branch for the short term.

sdk/Javascript/src/Microsoft.WindowsAzure.Mobile.JS.csproj
```diff
@@ -1,13 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <ToolsPath>$(MSBuildProjectDirectory)\..\tools</ToolsPath>
     <MakePriPath Condition="Exists('$(MSBuildProgramFiles32)\Windows Kits\8.0\bin\x64\makepri.exe')">$(MSBuildProgramFiles32)\Windows Kits\8.0\bin\x64\makepri.exe</MakePriPath>
     <MakePriPath Condition="Exists('..\..\Tools\makepri.exe')">..\..\Tools\makepri.exe</MakePriPath>
+    <TargetPlatformVersion>8.1</TargetPlatformVersion>
```
phvannor: What does this drive, I'm not sure I want to say no to 8.0 for Source, only for testing.

shrishrirang: Yeah right, will fix this when I update the PR for the new branch.

shrishrirang: Let us continue using this PR for review, I will create an 'offline' branch and send a separate PR for it once the code review is complete.

sdk/Javascript/src/MobileServiceSyncContext.js
```diff
@@ -0,0 +1,70 @@
...
+
+    this.client = client;
+
+    this.initialize = function (store) {
+        /// <summary>
+        /// Initializes the sync context with an instance of the store to be used
+        /// </summary>
+
+        if (store === undefined) {
+            this.store = new MobileServiceMemoryStore();
```
phvannor: the store should not be assumed by the SDK proper, it is a required param

sdk/Javascript/src/MobileServiceMemoryStore.js
```diff
@@ -0,0 +1,104 @@
...
+        /// <param name="instance" type="Object">
+        /// The object to insert into the table.
+        /// </param>
+        /// <param name="callback" type="Function">
+        /// The callback to invoke when the insert operation is complete, either successfully or unsuccessfully
+        /// </param>
+
+        // Platform.async will invoke this method by passing a callback object as the last argument
+        // Make sure that the callback object passed is same as the callback argument to avoid not fulfilling the promise ever.
+        Validate.length(arguments, 3, 'arguments');
```
phvannor: callback should be optional

sdk/Javascript/src/MobileServiceMemoryStore.js
```diff
@@ -0,0 +1,104 @@
...
+// ----------------------------------------------------------------------------
+
+/// <reference path="C:\Program Files (x86)\Microsoft SDKs\Windows\v8.0\ExtensionSDKs\Microsoft.WinJS.1.0\1.0\DesignTime\CommonConfiguration\Neutral\Microsoft.WinJS.1.0\js\base.js" />
+/// <reference path="C:\Program Files (x86)\Microsoft SDKs\Windows\v8.0\ExtensionSDKs\Microsoft.WinJS.1.0\1.0\DesignTime\CommonConfiguration\Neutral\Microsoft.WinJS.1.0\js\ui.js" />
+/// <reference path="Generated\MobileServices.DevIntellisense.js" />
+
+var Validate = require('Validate');
+var Platform = require('Platform');
+
+function MobileServiceMemoryStore() {
```
phvannor: this should go in test.

sdk/Javascript/src/MobileServiceMemoryStore.js
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// Initializes a new instance of the MobileServiceMemoryStore class.
+    /// </summary>
+
+    var idProperty = "id";
+    var tables = {};
+
+    this.insert = Platform.async(function (tableName, instance, callback) {
```
phvannor: Store has upsert only, not insert/update

sdk/Javascript/src/MobileServiceMemoryStore.js
```diff
@@ -0,0 +1,104 @@
...
+
+        // To keep things simple, just verify that the table has an "id" property
+        Validate.notNull(instance[idProperty]);
+
+        var table = tables[tableName] = tables[tableName] || {};
+
+        var instanceId = instance[idProperty];
+        Validate.notNull(instanceId);
+
+        if (table[instanceId] !== undefined) {
```
phvannor: see above, this is not a case for the local store to worry about

sdk/Javascript/src/MobileServiceMemoryStore.js
```diff
@@ -0,0 +1,104 @@
...
+        /// </param>
+
+        // Platform.async will invoke this method by passing a callback object as the last argument
+        // Make sure that the callback object passed is same as the callback argument to avoid not fulfilling the promise ever.
+        Validate.length(arguments, 3, 'arguments');
+
+        Validate.isString(tableName, 'tableName');
+        Validate.notNullOrEmpty(tableName, 'tableName');
+
+        Validate.isString(id, 'id');
```
phvannor: this is theory shouldn't be required either, we should try to let the store deal with any Id type it can handle.

sdk/Javascript/src/MobileServiceSyncContext.js
phvannor: this.store is not defined anywhere, also since private should be _store, _client, etc

sdk/Javascript/src/MobileServiceMemoryStore.js
shrishrirang: OK, is there a reason for just having upsert and no insert?

sdk/Javascript/src/MobileServiceMemoryStore.js
shrishrirang: Yes, that is what I plan to do for SQL store which I am working on. DefineTable can specify the id column name. I have kept MobileServiceMemoryStore simple for now by assuming the id column will be named 'id'.

sdk/Javascript/src/MobileServiceMemoryStore.js
phvannor: Yes, this is a purposeful design decision on the local store interface.  Update is not usable from the local store layer (the SDK never knows if a record should be inserted or updated locally), and while we could in theory know when something is a pure insert (from a sync table insert) it didn't add much and so to keep the interface simple we left it out.

sdk/Javascript/src/MobileServiceMemoryStore.js
phvannor: I meant the data type, not the column name.  I wouldn't worry about column names changing at this point in the design.  Our local tracking tables may use int Ids for internal tables so we shouldn't force string on Id.

shrishrirang:
Closing this PR. Split this PR into 2 separate PRs: Azure/azure-mobile-services-test#89 and #777 - one for the JS SDK and one for the PhoneGap test changes.

PR #793, opened 2015-09-14
Added MobileServiceTableOperationError.UpdateOperationAsync

soninaren:


sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -303,6 +303,28 @@ public Task CancelAndUpdateItemAsync(MobileServiceTableOperationError error, JOb
             });
         }
 
+        public Task UpdateOperationAsync(MobileServiceTableOperationError error, JObject item)
+        {
+            string itemId = error.Item.Value<string>(MobileServiceSystemColumns.Id);
```
phvannor: Is error.item guaranteed to be populated, I thought it was optional.  (For instance a synced delete could be done on an item not present locally, etc)

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -303,6 +303,28 @@ public Task CancelAndUpdateItemAsync(MobileServiceTableOperationError error, JOb
...
+            return this.ExecuteOperationSafeAsync(itemId, error.TableName, async () =>
+            {
+                await this.TryUpdateOperation(error, item);
+                if (error.OperationKind != MobileServiceTableOperationKind.Delete)
```
phvannor: nit: use { }

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -303,6 +303,28 @@ public Task CancelAndUpdateItemAsync(MobileServiceTableOperationError error, JOb
...
+                throw new InvalidOperationException("The operation has been updated and cannot be updated again");
+            }
+
+            // delete errors for updated operation
+            await this.Store.DeleteAsync(MobileServiceLocalSystemTables.SyncErrors, error.Id);
+        }
+
         public Task CancelAndDiscardItemAsync(MobileServiceTableOperationError error)
         {
             string itemId = error.Item.Value<string>(MobileServiceSystemColumns.Id);
```
phvannor: I guess I should complain about this too then.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -303,6 +303,28 @@ public Task CancelAndUpdateItemAsync(MobileServiceTableOperationError error, JOb
...
+            {
+                await this.TryUpdateOperation(error, item);
+                if (error.OperationKind != MobileServiceTableOperationKind.Delete)
+                    await this.Store.UpsertAsync(error.TableName, item, fromServer: true);
+            });
+        }
+
+        private async Task TryUpdateOperation(MobileServiceTableOperationError error, JObject item)
+        {
+            if (!await this.opQueue.UpdateAsync(error.Id, error.OperationVersion, item))
```
phvannor: I didn't realize that error.id == operation.id, can we update error's help text to say that.  As is, it implies it unique string...

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
```diff
@@ -149,6 +149,32 @@ public Task<IDisposable> LockItemAsync(string id, CancellationToken cancellation
...
+            try
+            {
+                MobileServiceTableOperation op = await GetOperationAsync(id);
+                if (op == null || op.Version != version)
+                {
+                    return false;
+                }
+
+                // Change the operation state back to pending since this is a newly updated operation without any conflicts
+                op.State = MobileServiceTableOperationState.Pending;
```
phvannor: is this what we do already during a collapse, it doesn't feel right to me, so wanted to double check.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
```diff
@@ -149,6 +149,32 @@ public Task<IDisposable> LockItemAsync(string id, CancellationToken cancellation
...
+                
+                // if the operation type is delete then set the item property in the Operation table
+                if (op.Kind == MobileServiceTableOperationKind.Delete)
```
phvannor: nit: { }

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
```diff
@@ -149,6 +149,32 @@ public Task<IDisposable> LockItemAsync(string id, CancellationToken cancellation
...
+                    op.Item = item;
```
phvannor: else { op.Item = null }, since we aren't enforcing any old type to new type better be safe.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
```diff
@@ -149,6 +149,32 @@ public Task<IDisposable> LockItemAsync(string id, CancellationToken cancellation
...
 
+        public virtual async Task<bool> UpdateAsync(string id, long version, JObject item)
+        {
+            try
+            {
+                MobileServiceTableOperation op = await GetOperationAsync(id);
+                if (op == null || op.Version != version)
+                {
+                    return false;
+                }
```
phvannor: dont we need to inc version here as well?

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
```diff
@@ -303,6 +303,28 @@ public Task CancelAndUpdateItemAsync(MobileServiceTableOperationError error, JOb
             });
         }
 
+        public Task UpdateOperationAsync(MobileServiceTableOperationError error, JObject item)
+        {
+            string itemId = error.Item.Value<string>(MobileServiceSystemColumns.Id);
+            return this.ExecuteOperationSafeAsync(itemId, error.TableName, async () =>
```
phvannor: How does this play out being called from the delegate?  Can we add a test case or two for those flows?

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
soninaren: Checked all the code paths, error.item is populated from operation.I followed code pattern for cancelAndUpdate() and cancelAndDiscard() and also ran through all the scenarios for conflicts to verify that it is always populated.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
soninaren: will update the help text for error

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
soninaren: yes your are right, the operation version should be incremented here.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/Queue/OperationQueue.cs
soninaren: This can be discussed. Since the related errors have been cleared and operation has been essentially not attempted, it seemed logical to reset it to pending.

phvannor: The original idea of this flag, was is it possible the server has seen part of this request.  In this case, the answer to that is clearly Yes (since this is a conflict/etc).  I can see both ways here, it should be documented clearly that this resets our tracking of the op in the public docs on the method.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
soninaren: Done.

sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Table/Sync/MobileServiceSyncContext.cs
soninaren: https://github.com/soninaren/azure-mobile-services/blob/dev/sdk/Managed/test/Microsoft.WindowsAzure.MobileServices.Test/UnitTests/Table/Sync/MobileServiceSyncContext.Test.cs#l214-l257

Issue #794, opened 2015-09-16
Xcode 7 Project WindowsAzureMobileServices.framework bitcode enabled


tahneesmith:
I am trying to update my iOS app to make it compatible with Xcode 7/iOS 9 and I get this build error:

WindowsAzureMobileServices.framework/WindowsAzureMobileServices(MSDateOffset.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

The WindowsAzureMobileServices framework needs to be rebuilt with bitcode enabled in the build settings. 

This is extremely urgent...please update your iOS framework as soon as possible.

phvannor:
We will take a look at getting a build & release done. In the meantime, you can clone this repo and should be able to run bash build.command and get a framework generated for you.  (After making the necessary edit to the proj)

tahneesmith:
Thanks!

PR #796, opened 2015-09-17
[iOS] Update build process to include BitCode, fix include issue

phvannor:
Updates the build command to include BitCode in the generated framework

Fixes an include issue in MSCoreDataStore that stopped it from being used via a cocoapod


phvannor:
Fixes #794 by passing right flags to bash command.

PR #802, opened 2015-09-21
Generic types in objective-c available in Xcode 7

damienpontifex:
Implementing Objective-C generics for collections available with Xcode 7 compiler. Helpful for compiler checking in Objective-C but assists with stronger type constraints in Swift.

This will then mean the latest version of Xcode (currently 7.0 as of September 2015) is needed to compile, but as it's just a compiler feature, is still backwards compatible with iOS versions.


azurecla:

TTYL, AZPRBOT;

phvannor:
To support BitCode, I've already enforced XCode7 support shortly, so I'm ok with the new requirement.

sdk/iOS/src/MSQueryResult.h
```diff
@@ -10,7 +10,7 @@
 ///@{
 
 /// The result from a successful read
-@property (nonatomic, strong, readonly)           NSArray *items;
+@property (nonatomic, strong, readonly, nonnull)           NSArray<NSDictionary *> *items;
```
phvannor: I'm not sure on items having to be non null, logically it probably should be, but I dont think the code base always gives an array depending on the server response.  Did you confirm this?

phvannor: granted perhaps we should add some errors on table reads who don't return arrays that we can parse, as I think non arrays are what result in nil...

damienpontifex: There is an example of it being constructed from a `MSSyncContextReadResult`. The `MSSyncContextReadResult` has nonnull items property, so I propagated that through to this.

Haven't found a pathway where it is constructed with a nil property

damienpontifex:
In this block here https://github.com/Azure/azure-mobile-services/blob/MobileApp/sdk/iOS/src/MSTableConnection.m#L142-L162 there is existing code where `MSQueryResult` will potentially be instantiated with a nil value for the items array so changed the nullable annotation on this property

sdk/iOS/src/MSTableConnection.m
```diff
@@ -256,6 +256,39 @@ -(NSInteger) items:(NSArray **)items
     return totalCount;
 }
 
++(void) removeSystemColumnsFromItem:(NSMutableDictionary *)item ifNotInQuery:(NSString *)query
```
phvannor: This function should be gone in MobileApp, not sure why it got added again, can you drop this file's changes from yoru commit

PR #809, opened 2015-09-23
[Cordova] First set of changes for offline sync

shrishrirang:
- Added a basic implementation of SQLite store based on litehelpers' SQLite
- Some TODOs are marked in code and needs to be taken care of.
- More UTs need to be added. This commit contains a single UT to test basic functionality.
- Tweaked the test framework adapter to start running the tests only when asked to do so. This makes it possible to trigger the tests when all required cordova plugins are loaded. WinRT implementation already does this
- QUnit adds new elements to the DOM which is not allowed in WinRT unless done in an 'unsafe' context. Implemented a hack to workaround this limitation without having to edit QUnit code. This implementation makes an implicit assumption on the QUnit code (that the only disallowed calls are the calls to the toolbar's appendChild() method)
- Ported the existing browser UTs to Cordova. Fixed some incorrect platform checks (isWinRT, etc).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,5 @@ packages
 *.jar
 *.xccheckout
 deriveddata
-
+.vs
 *.iml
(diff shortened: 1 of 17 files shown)


shrishrirang:
Ping!

PR #815, opened 2015-10-07
[Javascript] Use browserify for bundling JS modules

shrishrirang:
- Use browserify to bundle the modules instead of the home-grown custom bundling tool
- Remove submodules esprima and queryjs and use them as npm modules instead
- Remove custom code to expose modules by adding them to global namespace (InternalsVisible.js).
- The '*.Internals.js' bundles now expose modules needed by the UTs (Platform, Validate and Extensions)
- Platform is defined as an alias. It will map either to the web or winjs implementation depending on how it is defined at build time. Alias is used as conditional 'require' has limitations with browserify.
- Moved Platform.*.js/MobileSQLiteStore.js to platform specific directories
- Remove unnecessary entry points from Gruntfile.js
- Remove nodeexports.js as it is not needed anymore


shrishrirang:
Ping

PR #818, opened 2015-10-13
Fix the broken links for iOS Reference Docs

manimaranm7:
Turns out the links are case sensitive, so changed part of the hyperlink from "ios" to "iOS"

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ To get the source code of our SDKs and samples via **git** just type:
 ## Reference Documentation
 
 #### Azure App Service Mobile Apps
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@manimaranm7__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no contribution license agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, AZPRBOT;

PR #822, opened 2015-10-14
[JS] Send 2.0 API Version header

phvannor:


sdk/Javascript/src/MobileServiceClient.js
```diff
@@ -231,6 +231,7 @@ MobileServiceClient.prototype._request = function (method, uriFragment, content,
     if (!_.isNullOrEmpty["X-ZUMO-VERSION"]) {
         options.headers["X-ZUMO-VERSION"] = this.version;
     }
+    options.headers["ZUMO-API-VERSION"] = "2.0.0";
```
shrishrirang: Platform.getSdkInfo().fileVersion instead of 2.0.0?

phvannor: That would be the SDK version, this is the new API version that is independent of the specific client bits, just what protocol it uses.  More details in the team one note.

shrishrirang: Got it!

PR #830, opened 2015-10-28
[iOS] Easy Auth Support

phvannor:


sdk/iOS/src/MSClient.m
```diff
@@ -67,90 +67,82 @@ -(NSString *) installId
...
-    self = [super init];
-    if(self)
-    {
-        _applicationURL = applicationUrl;
-        _gatewayURL = gatewayUrl;
-        _applicationKey = [key copy];
-        _login = [[MSLogin alloc] initWithClient:self];
-        _push = [[MSPush alloc] initWithClient:self];
+    if (!_loginPrefix) {
+        _loginPrefix = @".auth/login";
```
pragnagopa: For managed prefix has a leading slah: /.auth/login for login url to point at host level. If MobileAppUrl: https://a/b, login url: https://a/.auth/login. Will it be the same here?

phvannor: Yes, appendByPathComponent, will add a slash if neither the original url or the path components has a slash. If one of them has it, then it will not.

sdk/iOS/test/MSClientTests.m
```diff
@@ -145,27 +101,58 @@ -(void) testInitWithApplicationURLAllowsNilURL
...
+
+    client.loginPrefix = @".auth/login";
+    XCTAssertEqualObjects(client.loginPrefix, @".auth/login");
     
-    XCTAssertNotNil(client.applicationKey, @"client.applicationKey should not be nil.");
-    XCTAssertTrue([client.applicationKey isEqualToString:@"here is some key"],
-                 @"The client should be using the url it was created with.");
+    client.loginPrefix = @".auth/ space/!@#";
+    XCTAssertEqualObjects(client.loginPrefix, @".auth/%20space/!@%23");
 }
```
pragnagopa: Add another test where both host and prefix are set

phvannor: updated structure slightly so loginURL can be tested this way and see how they combine in the test since Login and Controller aren't unit tested atm.

Issue #680, opened 2015-04-21
Inc Sync: Use updatedAt, Id as sort


phvannor:
Revamp inc sync logic to use a secondary key of Id for the sorting.  This would allow the paging to resume exactly where it left off.  *see: Android implementation for details.

paulbatum:
We probably won't backport this change but we should consider it for the new version of the client. Filed Azure/azure-mobile-apps-net-client#51

PR #856, opened 2015-11-19
Incremental Update fetches all items

mikaelbartlett:
When using sync tables and query id to get an incremental update the requests filters with greater than or equal witch results in all items being fetched again. Using iOS.

$skip=0&$filter=(__updatedAt%20ge%20datetimeoffset'2015-10-28T14%3A04%3A10.590Z')&$orderby=__updatedAt%20asc&__includeDeleted=true&$top=50&__systemProperties=__createdAt%2C__updatedAt%2C__deleted%2C__version

### DIFF
--- a/sdk/iOS/src/MSQueuePullOperation.m
+++ b/sdk/iOS/src/MSQueuePullOperation.m
@@ -305,7 +305,7 @@
     
     if (self.deltaToken) {
         MSDateOffset *offset = [[MSDateOffset alloc]initWithDate:self.deltaToken];
-        NSPredicate *updatedAt = [NSPredicate predicateWithFormat:@"%K >= %@", MSSystemColumnUpdatedAt, offset];
+        NSPredicate *updatedAt = [NSPredicate predicateWithFormat:@"%K > %@", MSSystemColumnUpdatedAt, offset];
         if (self.originalPredicate) {
             self.query.predicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[self.originalPredicate, updatedAt]];
         }


azurecla:
Hi __@mikaelbartlett__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no contribution license agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, AZPRBOT;

azurecla:
__@mikaelbartlett__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

phvannor:
@mikaelbartlett this is a slight limitation of the current design.  

The use of >= is do to potentially stopping in the middle of paging.  Say you had 80 records of time X, and your first pull gets 50, then the next one would use a $skip=50 to get the next set.  However if you lose connection we have to stop.

With this proposed fix, starting from > when we resumed again, we'd miss the remaining 30 we need to pick up.

With >=, we'd get all 80 again.  One common proposal then is to cache the skip (50). We dont do that as any of the 50 records at time X may have been altered.  (this is technically always a slight risk in a highly active table, but active tables tend not to have big blocks of records with same timestamp)

As an example, say record 10 was updated while you were offline by another client, now only 49 records that you got are still at time X, so skipping 50 would miss the original record 50.

If we look here: https://github.com/Azure/azure-mobile-services/issues/680 we see that the current proposed fix here is to sort by a secondary key (id), and then change the filter from updatedAt >= X to (updated = X && id > lastIdFromPrevPull) || updatedAt > X)

It looks like Android has a bug, or it took an optimization I'm not quite familiar with to drop the other clause.  (Maybe if it ran to the very end & got 0 results from the server you can drop the extra clause?)

phvannor:
see comments above, logic is currently intended (if a lil extra work in some cases)

PR #863, opened 2015-12-16
Backporting fixes to properly work with Xamarin linker

### LINKED ISSUE #855: In MobileServiceHttpClient class, remove the AssemblyFileVersion from the building of user agent string
In Xamarin.iOS, (where I believe this library is very commonly used) the linker strips away some `AssemblyInfo` attributes, including `AssemblyFileVersionInfo`. The value from this attribute is used [in the bowels of the MobileServiceHttpClient](https://github.com/Azure/azure-mobile-services/blob/master/sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Http/MobileServiceHttpClient.cs#L695), where it's used to buildup the user agent string; I presume so that MSFT can measure what versions of their client SDK are hitting their endpoints.

The problem is that when linking is enabled in a Xamarin.iOS solution (and presumably in other linkers as well), the `AssemblyFileVersionInfo` attribute is stripped off. So the result of the LINQ statement on line 695 is null. And then on the very next line, the method attempts to access the `Version ` property on the null `fileVersionAttribute` local variable.

I understand the reason for trying to grab the value: to track client SDK versions. But basing it on an attribute that gets stripped off with linking probably isn't a very good way to achieve it. Just my two cents! Thanks!

fabiocav:
Backporting fixes to properly work with Xamarin linker: https://github.com/Azure/azure-mobile-apps-net-client/pull/9

Resolves #855


fabiocav:
@paulbatum / @phvannor this is the same work done in #855 , backported to Mobile Services.

Issue #844, opened 2015-11-11
[iOS] Exception in Push.RegisterNativeAsync method


AlexS01:
When use library on iOS with Xamarin.Forms.

```
public class AzureNotificationHubService : IAzureNotificationHubService
{
    private readonly MobileServiceClient _mobileServiceClient;
    private readonly Push _push;

    public string CurrentDeviceId { get; private set; }

    public AzureNotificationHubService()
    {
        _mobileServiceClient = new MobileServiceClient(AzureMobileService.ApplicationURL, AzureMobileService.ApplicationKey);
        _push = _mobileServiceClient.GetPush();            
    }

    public async Task UnregisterNativeAsync()
    {
        await _push.UnregisterNativeAsync();
    }

    public async Task RegisterNativeAsync(string deviceId)
    {
        CurrentDeviceId = deviceId;
        
        try
        {
            await _push.RegisterNativeAsync(deviceId);
        }
        catch (Exception exception)
        {
            // EXCEPTION THERE
        }
    }
}
```

Nov 11 08:45:10 SOME-iPad XXXappiOS[1387] <Warning>: 11/11/15 8:45:10 - *** Push Notification - Device Registered - Token : 29b5b1bd03f1c22939e5515c388f7258964ba180393960dd849fe10f844eb1fd
Nov 11 08:45:11 SOME-iPad XXXappiOS[1387] <Warning>: 11/11/15 8:45:11 - *** RegisterNativeAsync exception
Nov 11 08:45:11 SOME-iPad XXXappiOS[1387] <Warning>: 11/11/15 8:45:11 - Source: mscorlib;
Nov 11 08:45:11 SOME-iPad XXXappiOS[1387] <Warning>: Message:Argument cannot be null.
Nov 11 08:45:11 SOME-iPad XXXappiOS[1387] <Warning>: Parameter name: key
Nov 11 08:45:11 SOME-iPad XXXappiOS[1387] <Warning>: StackTrace:  at System.ThrowHelper.ThrowArgumentNullException (ExceptionArgument argument) [0x00000] in /Users/builder/data/lanes/2320/bae2cdbe/source/maccore/_build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/src/mono/external/referencesource/mscorlib/system/throwhelper.cs:82
[...]

PR #864, opened 2015-12-21
Use default registration name in LocalStoreManager

pragnagopa:
Fix for issue #844


sdk/Managed/src/Microsoft.WindowsAzure.MobileServices/Push/LocalStorageManager.cs
```diff
@@ -96,7 +96,10 @@ public void DeleteRegistrationByRegistrationId(string registrationId)
         public void UpdateRegistrationByName(string registrationName, string registrationId, string registrationDeviceId)
         {
             StoredRegistrationEntry cacheReg = new StoredRegistrationEntry(registrationName, registrationId);
-
+            if (string.IsNullOrEmpty(registrationName))
```
phvannor: was there a unit test capturing this already?  If not, can you add one?

pragnagopa: Added unit tests for LocalStorageManager

phvannor:
:+1:

